PR #12, opened 2021-09-17
Set up a very basic pandoc markdown file + Makefile

kbeyls:


g-kouv (approved):

g-kouv (approved):

Makefile
```diff
@@ -2,5 +2,10 @@ PANDOCFLAGS = \
 	      --table-of-contents \
 	      --number-sections
 
-build/book.pdf: book.md Makefile
+all: build/book.pdf
```
g-kouv, line 5: Perhaps we should also have:
.PHONY: all
?

kbeyls: That makes sense, thank you!
I've added the .PHONY line just now in the latest update.

PR #13, opened 2021-09-21
Add docker container to build the book.

kbeyls:
This makes it easy to make sure that the book is built in a reproducible
way on different machines.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,0 +1,24 @@
+# syntax=docker/dockerfile:1
+FROM ubuntu:20.04
+# install pandoc and texlive dependencies
+RUN apt-get update && apt-get install -y wget perl pandoc make\
+    && rm -rf /var/lib/apt/lists/*
+# Install from stable Texlive 2020 final repo
+RUN cd /root && \
+    wget https://ftp.tu-chemnitz.de/pub/tug/historic/systems/texlive/2020/install-tl-unx.tar.gz && \
+    mkdir install-tl-unx && \
+    tar -zxvf install-tl-unx.tar.gz -C install-tl-unx --strip-components=1 && \
+    echo "selected_scheme scheme-minimal" >> texlive.profile && \
+    mkdir texlive && \
+    TEXLIVE_INSTALL_PREFIX=/texlive ./install-tl-unx/install-tl \
+        -profile texlive.profile \
+        -repository https://ftp.tu-chemnitz.de/pub/tug/historic/systems/texlive/2020/tlnet-final && \
+    ln -s /texlive/2020/bin/* /texlive/bin
+ENV PATH=/texlive/bin:$PATH
+# install latex too (the minimal scheme selected above does not install latex)
+RUN tlmgr install latex latex-bin
+# packages needed for pandoc:
+RUN tlmgr install amsfonts amsmath lm unicode-math iftex listings fancyvrb booktabs graphics hyperref xcolor ulem geometry setspace babel upquote microtype parskip xurl bookmark footnotehyper mdwtools kvoptions etoolbox pdftexcmds infwarerr
+
+COPY entrypoint.sh /entrypoint.sh
+ENTRYPOINT ["/entrypoint.sh"]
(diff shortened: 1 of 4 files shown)


docker/Dockerfile
```diff
@@ -0,0 +1,24 @@
+# syntax=docker/dockerfile:1
+FROM ubuntu:20.04
+# install pandoc and texlive dependencies
+RUN apt-get update && apt-get install -y wget perl pandoc make\
+    && rm -rf /var/lib/apt/lists/*
+# Install from stable Texlive 2020 final repo
```
fpetrogalli, line 6: Any particular reason for not using the version of texlive shipped with the distro? if so, maybe a comment explaining why you need to get texlive from the tex server?

To me lines 7-21 could have been replaced with `RUN apt-get install texlive-latex-extra`.

Also, there are dockers already available with pandoc + latex in it: https://pandoc.org/installing.html#docker

Maybe you could simplify the machinery with the instruction provided there?

kbeyls: Many thanks for having a look @fpetrogalli !

There are a few reasons why I'm not going for the texlive packages as provided by the linux distro:
1. It results in a (much) bigger docker container. By going for a minimal texlive installation, just installing what is needed, the docker container can remain significantly smaller. Having it measured just now, it's 572MB versus 1.03GB...
2. In later pull requests, I plan to use a few more latex packages that will have to be installed tlmgr (cannot be installed using apt-get), so the docker container will need tlmgr functionality anyway.

There are indeed docker containers available elsewhere, but with those containers, I guessed there may be less control over exactly which versions of latex or pandoc would be used. Or at least be harder to figure out which versions are used.

All those reasons together, it seemed to me it was worthwhile, and ultimately simpler, to just install directly using texlive.

If the machinery seems too complex, I could add extra comments. Are there specific parts that remain unclear to you with the existing comments?

fpetrogalli: Yeah, the machinery is complex and deserves a couple of comments explaining why you prefer the texlive installation over the system one.

However:

1. How important it is for you to minimize the size of the docker image? I believe (hope) that the CI infrastructure will have most of these images cached/stored in some local mirror? Even for the local builds... contributors will contribute mostly to the book, not to the docker image, so once they have done the first build of it they will likely not have to change it? (Admittedly, this can be seen also as an argument to leave the dockerfile as complex as it is... :) )
2. The docker containers available with preinstalled docker and latex have tags that can prevent version changes of the tools you use to build the book: https://hub.docker.com/r/pandoc/latex/tags?page=1&ordering=last_updated If you rely on those, you don't have to worry about accidentally changing the outcome of the book because of an upgrade
3. Re: installing more latex packages than what is available in the distribution. `tlmgr` is not available with the system installation of texlive via apt-get? It seems to be part of `texlive-base`: http://manpages.ubuntu.com/manpages/trusty/man1/tlmgr.1.html

All in all, I would still simply go with the docker image. I won't hold my position here though, it is up to you to decide what you think is the best solution for the project! Just add some explanation on the reasoning for a manual installation of texlive in the comments :)

Ciao,

Francesco

kbeyls: I agree that the trade-offs between using apt-get texlive and tlmgr are minor, hence both options are probably fine.
Trying to share a few more of my thoughts based on your questions:
1. "How important it is for you to minimize the size of the docker image?" Indeed, for local builds this probably will not matter much as the image will be cached/stored locally. However, for CI builds (such as using github actions), my understanding is that either the image always gets recreated from scratch; or if extra work is done to enable caching, the image will be downloaded over a network link. Saving 500MB seems like a measurable saving in that context. The reason in CI setups like github actions that this cannot be cached/stored locally is that they run in VMs.
2. Agreed, relying on tacker docker hub containers means we have stable containers. It's just somewhat less visible exactly which version of texlive and other packages we exactly rely upon. Not sure if that matters too much. I guess my personal preference is slightly more towards explicitly seeing the root dependencies?
3. Indeed tlmgr is installed if you do install texlive via apt-get. Again, it's probably just personal taste to use only one system for everything TeX related, i.e. the texlive system, rather than 2 (the apt-get/ubuntu repositories + the tlmgr TeXlive package system).

Thank you very much for sharing your thoughts.
It seems it highlights that both options are probably equally fine.
If in the future it turns out that's not the case, at least we'll have recorded in this conversation what the trade-offs are we thought of right now :)

Thanks!

Kristof

fpetrogalli: LGTM! Thank you for your patience!

fpetrogalli (approved):

kbeyls:
@all-contributors please add @fpetrogalli for review

PR #17, opened 2021-09-27
Add ability to add items to index.

kbeyls:
This is based on latex makeindex functionality.
Adding a word to the index in the markdown text is as simply as
writing \index{word-to-add-to-index}.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,12 +1,22 @@
+# SPDX-License-Identifier: CC-BY-4.0
+# SPDX-FileCopyrightText: © 2021 Arm Limited <kristof.beyls@arm.com>
+
 PANDOCFLAGS = \
 	      --table-of-contents \
-	      --number-sections
+	      --number-sections \
+	      --standalone
 
-.PHONY: all
+.PHONY: all clean
 all: build/book.pdf
+clean:
+	rm -rf build
 
 build:
 	mkdir build
 
-build/book.pdf: book.md Makefile build
-	pandoc $< -o $@ $(PANDOCFLAGS)
+build/book.tex: book.md Makefile build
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -1,12 +1,24 @@
...
+build/book.tex: book.md Makefile build
+	pandoc $< -t latex -o $@ $(PANDOCFLAGS)
+
+build/book.pdf: build/book.tex Makefile build
+	cd build && \
+	pdflatex book.tex -o book.pdf $(PANDOCFLAGS) && \
+	pdflatex book.tex -o book.pdf $(PANDOCFLAGS) && \
+	makeindex book.idx && \
+	pdflatex book.tex -o book.pdf $(PANDOCFLAGS) && \
+	pdflatex book.tex -o book.pdf $(PANDOCFLAGS)build/book.pdf: book.md Makefile build
```
g-kouv, line 23: This looks odd. Copy-paste issue?

kbeyls: That indeed looks like a copy paste error.
Looking at it now, those $(PANDOCFLAGS) also shouldn't be there on the pdflatex command line.

Maybe in the end, rather than invoking pdflatex and makeindex multiple times manually, maybe it'd be better to use https://mg.readthedocs.io/latexmk.html.

I also noticed now that the Makefile is missing a copyright header....

I hope to have a new version of the PR up later today with the above fixed.

kbeyls: I've just pushed an updated commit on this PR to address all of the above.
I decided to switch to using latexmk; that feels less of a hack than invoking pdflatex multiple times (without necessarily being sure that it is invoked enough).

g-kouv (approved):

PR #18, opened 2021-09-28
Add (empty) bibliography.

kbeyls:


kbeyls:
It turns out that the version of pandoc that ships with Ubuntu 20.04 (and also the one that ships with most recent Ubuntu 21.04) is too old to understand the --citeproc command. Therefore, I've amended the commit in this PR to use the older style '--filter pandoc-citeproc' option

g-kouv (approved):

PR #19, opened 2021-09-28
Introduce the ability to add todos inline in the document

kbeyls:
The idea is that this makes it possible to develop content for the book using a similar trick as with software development where one sometimes writes "FIXME" or "TODO" comments to not get mentally stuck on a detail.
The hope is that having a simple way to jot down TODOs while writing content helps to make a note that more detail could be added to a section; while not holding up getting a section into review before that detail is added.

### DIFF
--- a/book.md
+++ b/book.md
@@ -10,6 +10,9 @@ header-includes:
   ```{=latex}
   \usepackage{makeidx}
   \makeindex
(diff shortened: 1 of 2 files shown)


g-kouv (approved):
The TODOs look very nice indeed!

PR #24, opened 2021-09-29
Add timing attacks section

kbeyls:


statham-arm:
One thing I'd add to this: cache side channels.

Direct timing side channels of the kind you describe, where the attacker measures the run time of the actual secret-handling code, are only part of the story. Another important thing is to avoid leaving traces in the cache that are correlated to your secret data, because those can be observed by measuring the run time of further code running afterwards, and perhaps even probed on purpose by sending appropriate inputs.

When I hardened PuTTY's crypto code against these kinds of side channel attack, I worked to three principles. Two of them you've got already (don't vary control flow based on secrets, and avoid arithmetic operations whose data-independence you're not confident of). The third is: don't vary your sequence of _memory accesses_ based on secrets. For example, this is why you shouldn't use an in-memory lookup table to implement the AES S-box.

Secondly, perhaps it's worth discussing some typical techniques for coding to these constraints? I for one always find it annoying to be told 'Don't do this!' without the slightest hint about what I _should_ do instead.

An important trick is bitwise selection, for example. Suppose you have a boolean flag `x`, in the form of an integer that's either 0 or 1, and you want to set a result variable `r` to one of inputs `a` and `b` depending on the value of the flag. If you write `r = x ? a : b` or similar, the compiler may very well generate a conditional branch. (You _might_ get lucky and find you got a time-safe `CSEL` or some such, but you can't rely on it; and if your code is intended to be portable, then some of its target platforms probably won't even have an instruction of that kind.) A better approach is to start by converting your boolean flag into a whole word of bitmask that's all 0s or all 1s. Then use that to select between the values by writing something like `r = b ^ (bitmask & (a ^ b))`. So if the bitmask is all zeroes, you're XORing zero into `b`; and if it's all ones, you're computing `b ^ a ^ b`, i.e. `a`.

A similar bit-twiddling trick is used to create a safe memory compare (e.g. for checking a password or hash or credential without giving away where the first differing byte is). XOR each byte of `a` with the corresponding byte of `b`; AND together all of those bytes. That gives you an 8-bit value that is 0 if the whole memory regions are equal, and nonzero if they're not; then there are various bit-tricks you can use to normalise that to a nice clean output of either 0 or 1, such as adding 0xFF (which will set the 8th bit if and only if any bit below that was nonzero) and then shifting right by 8.

A particular principle to watch out for, if you're coding in C, is: beware of the C99 `bool` type. It looks harmless enough – just a 1-bit integer, right? – except that it has the feature that the implicit conversion from any other integer type to `bool` involves _any nonzero value_ being normalised to 1. How does the compiler code-generate that? Probably with a comparison and conditional branch! So in all the time-safe code I write, I revert to the old-school C approach of using an `unsigned` containing either 0 or 1 as my representation of a boolean, and I never use a C99 `bool` until I'm out of the time-critical section and into a less dangerous part of the code.

kbeyls:
@all-contributors please add @statham-arm  for review, ideas.

g-kouv (approved):

Issue #28, opened 2021-10-01
Also produce a nice-looking HTML version of the book.


kbeyls:
Currently, only a PDF version of the book is produced.

We should also produce a nice-looking HTML version of the book.
That is needed, at least, to enable #8.

It looks like out-of-the-box there may be limitations to in how far HTML output in pandoc can be customized using pandoc templates. Maybe we'll need to write a few pandoc filters to make the HTML output as nice as it can/should be?

PR #29, opened 2021-10-01
Also generate HTML output

kbeyls:
This is a draft series of commits enabling to also produce an HTML version of the book.
The goal for this pull request is to implement #28


kbeyls:
Abandoning this more complex approach in favor of the simpler approach in PR #30

PR #30, opened 2021-10-04
Produce a basic HTML version too.

kbeyls:
This uses the HTML produced by default by the pandoc framework.
For now, there is very little tuning done on the output, just
a few CSS rules so that the table of contents is displayed on
the left and the main text next to that.
The CSS also asks for a sans serif font, which is just a personal
preference.

This also updates the name of the script to build the outputs, as
it no longer produces only PDF.

The github action workflow is adapted to also archive the HTML and
CSS files produced.

The goal for this pull request is to implement #28


Makefile
```diff
@@ -6,15 +6,25 @@ PANDOCFLAGS = \
...
+pdf: build/book.pdf
+html: build/book.html build/default.css
 clean:
 	rm -rf build
 
 build:
 	mkdir build
 
-build/book.tex: book.md Makefile build
+build/default.css: default.css Makefile
```
g-kouv, line 19: Why does this depend on the Makefile?

g-kouv (approved):
Looks good for a first HTML version.

I'm guessing if we later (once more content is in place) want to have per-chapter pages, we will need to start keeping separate .md files per chapter?

kbeyls:
Yes, once we have more content, I'm pretty sure that the HTML version will have to be improved.
I hope that if we want to have separate web pages per chapter we won't be forced to have to separate the source .md files at the same time. Let's cross that bridge when we come to it.

PR #31, opened 2021-10-04
Automatically publish to github pages

kbeyls:
This is done by running a github action step publishing the entire build
directory to the gh-pages branch. This only done for build actions that
are run on the main branch, and only when triggered by a push (and not
e.g. by a pull request).

One other minor addition is to make the book.html also available under
index.html, so that that's the page that gets picked up automatically on
the github pages webpage.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,4 +1,4 @@
-name: Build book with docker container CI
+name: Build
 
 on:
   push:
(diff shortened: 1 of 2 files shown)


kbeyls:
This implements issue #8

g-kouv (approved):
Nice! So https://llsoftsec.github.io/llsoftsecbook/ will show the HTML version after this is merged, right?

kbeyls:
Indeed, the HTML version is now up at  https://llsoftsec.github.io/llsoftsecbook/ :)

PR #35, opened 2021-10-07
Remove unused group field from bibliography entry.

g-kouv:
This is to remove the unused "groups" field from the bibliography entry.


kbeyls (approved):
LGTM, thank you!

PR #43, opened 2021-10-08
Add introduction and section outline of memory vulnerabilities chapter

g-kouv:
A first version of the introduction to the memory vulnerabilities chapter. PTAL!

### DIFF
--- a/book.md
+++ b/book.md
@@ -109,52 +109,72 @@ feedback to be received through <https://github.com/llsoftsec/llsoftsecbook>.
  
 # Memory vulnerability based attacks and mitigations
 
-## Introduction
+## A bit of background on memory vulnerabilities
 
-An important class of vulnerabilities arise due to memory access errors, such as
-buffer overflows, use-after-free accesses and null pointer dereferences. An
-attacker can exploit such vulnerabilities to leak sensitive data or
-overwrite critical memory locations and gain control of the vulnerable program.
-These types of vulnerabilities most commonly occur due to programming errors when
-using languages such as C/C++, which do not provide mechanisms to protect the
-integrity of memory accesses by default.
+Memory access errors describe memory accesses that, although permitted by a
+program, were not intended by the programmer. These types of errors are usually
+defined [@Hicks2014] by explicitly listing their types, which include:
 
-Memory vulnerabilities have a long history. The Morris worm in 1988 was the
-first widely publicized attack exploiting a buffer overflow. Later, in the
-mid-90s a few famous write-ups describing buffer overflows appeared
-[@AlephOne1996].  Stack overflows were mitigated with stack canaries and
-non-executable stacks, but attackers answered with more ingenious ways to
-bypass these mitigations: code reuse attacks, starting from simple attacks like
-return-into-libc [@Solar1997] and leading up to Return-Oriented Programming
+  * buffer overflow
+  * null pointer dereference
+  * use after free
+  * use of uninitialized memory
+  * illegal free
+
+Memory vulnerabilities are an important class of vulnerabilities that arise due
+to these types of errors, and they most commonly occur due to programming
+mistakes when using languages such as C/C++. These languages do not provide
+mechanisms to protect against memory access errors by default. An attacker can
+exploit such vulnerabilities to leak sensitive data or overwrite critical
+memory locations and gain control of the vulnerable program.
+
+Memory vulnerabilities have a long history. The [Morris
+worm](https://en.wikipedia.org/wiki/Morris_worm) in 1988 was the first widely
+publicized attack exploiting a buffer overflow. Later, in the mid-90s, a few
+famous write-ups describing buffer overflows appeared [@AlephOne1996]. [Stack
+overflows](#stack-overflows) were mitigated with [stack
+canaries](#stack-overflows) and [non-executable stacks](#stack-overflows). The
+answer was more ingenious ways to bypass these mitigations: [code reuse
+attacks](#code-reuse-attacks), starting with attacks like
+[return-into-libc](#code-reuse-attacks) [@Solar1997]. Code reuse attacks later
+evolved to [Return-Oriented Programming (ROP)](#code-reuse-attacks)
 [@Shacham2007] and even more complex techniques.
 
-To defend against code reuse attacks, the defensive security side came up with
-address space layout randomization (ASLR) and control-flow integrity (CFI)
-measures. The race continues to this day. Each newly deployed mitigation
-results in attempts, often successful, to bypass it, or in alternative, more
-complex exploitation techniques, and even tools to automate them.
+To defend against code reuse attacks, the [Address Space Layout Randomization
+(ASLR)](#code-reuse-attacks) and [Control-Flow Integrity
+(CFI)](#code-reuse-attacks) measures were introduced. \todo{Refine section
+links used here and in the previous paragraph.} This interaction between
+offensive and defensive security research has been essential to improving
+security, and continues to this day. Each newly deployed mitigation results in
+attempts, often successful, to bypass it, or in alternative, more complex
+exploitation techniques, and even tools to automate them.
 
 Memory safe [@Hicks2014] languages are designed with prevention of such
 vulnerabilities in mind and use techniques such as bounds checking and
 automatic memory management. If these languages promise to eliminate
 memory vulnerabilities, why are we still discussing this topic?
 
-On one hand, C and C++ remain very popular languages, particular in the
+On the one hand, C and C++ remain very popular languages, particular in the
 implementation of low-level software. On the other hand, programs written in
 memory safe languages can themselves be vulnerable to memory errors as a result
-of bugs in their implementation. Can we fix the problem by switching to memory
-safe languages for compiler and runtime implementation? Even if that were as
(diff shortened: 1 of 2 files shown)


book.md
```diff
@@ -109,7 +109,73 @@ feedback to be received through <https://github.com/llsoftsec/llsoftsecbook>.
...
+
+An important class of vulnerabilities arise due to memory access errors, such as
+buffer overflows, use-after-free accesses and null pointer dereferences. An
+attacker can exploit such vulnerabilities to leak sensitive data or
+overwrite critical memory locations and gain control of the vulnerable program.
+These types of vulnerabilities most commonly occur due to programming errors when
+using languages such as C/C++, which do not provide mechanisms to protect the
+integrity of memory accesses by default.
+
+Memory vulnerabilities have a long history. The Morris worm in 1988 was the
```
kbeyls, line 122: Could we have a reference to the Morris worm?
maybe https://en.wikipedia.org/wiki/Morris_worm?
I wonder if we refer to wikipedia maybe we should just use a direct URL to it, rather than a citation?

book.md
```diff
@@ -109,7 +109,73 @@ feedback to be received through <https://github.com/llsoftsec/llsoftsecbook>.
...
+first widely publicized attack exploiting a buffer overflow. Later, in the
+mid-90s a few famous write-ups describing buffer overflows appeared
+[@AlephOne1996].  Stack overflows were mitigated with stack canaries and
+non-executable stacks, but attackers answered with more ingenious ways to
+bypass these mitigations: code reuse attacks, starting from simple attacks like
+return-into-libc [@Solar1997] and leading up to Return-Oriented Programming
+[@Shacham2007] and even more complex techniques.
```
kbeyls, line 129: In this sentence, a lot of new terms get introduced that will be explained in more detail later.
Would it be possible to somehow make it clear to the reader that it's OK to not understand what these terms mean here and that they will be explained later?
Maybe that could be done by making the terms links to the sections that explain them in more detail? (I'm not sure if that would work well?).
I think the following terms introduced here, and maybe in the next few sentences, could use that: stack overflows, stack canaries, non-executable stacks, code reuse attacks, return-to-libc, Return-Oriented Programming, ASLR, CFI. Maybe also "code reuse attacks"?

kbeyls, line 129: This is a longer sentence. Maybe it could be split into 2 sentences by ending the first sentence at "mitigations: code reuse attacks. Code reuse attacks have evolved from simple attacks like ...".

book.md
```diff
@@ -109,7 +109,73 @@ feedback to be received through <https://github.com/llsoftsec/llsoftsecbook>.
...
+
+To defend against code reuse attacks, the defensive security side came up with
```
kbeyls, line 131: "the defensive security side": this seems to be the first place that starts implicitly referring to there being an arm's race between a defensive and an offensive side. I wonder if this would be easier to consume if there was a paragraph before this one that explains there being such an arm's race?
Or maybe, the intent of this paragraph is exactly to introduce the arm's race? In that case, I think it may be better to start the paragraph more explicitly with something like "Over the past few decades, there's an arms race between attackers (the offensive side) and people implementing hardening techniques or mitigations (the defensive side)."

book.md
```diff
@@ -109,7 +109,73 @@ feedback to be received through <https://github.com/llsoftsec/llsoftsecbook>.
...
+address space layout randomization (ASLR) and control-flow integrity (CFI)
+measures. The race continues to this day. Each newly deployed mitigation
+results in attempts, often successful, to bypass it, or in alternative, more
+complex exploitation techniques, and even tools to automate them.
+
+Memory safe [@Hicks2014] languages are designed with prevention of such
+vulnerabilities in mind and use techniques such as bounds checking and
+automatic memory management. If these languages promise to eliminate
+memory vulnerabilities, why are we still discussing this topic?
```
kbeyls, line 140: I really like the "why are we still discussing this topic?" question here. Nice!

book.md
```diff
@@ -109,7 +109,73 @@ feedback to be received through <https://github.com/llsoftsec/llsoftsecbook>.
...
+
+On one hand, C and C++ remain very popular languages, particular in the
```
kbeyls, line 142: Minor minor style question: "On one hand" versus "On the one hand"?
I found this https://jakubmarian.com/on-the-one-hand-on-one-hand/ suggesting "On the one hand" might be preferred. Maybe best to check if there's anything about this in CMoS?

book.md
```diff
@@ -109,7 +109,73 @@ feedback to be received through <https://github.com/llsoftsec/llsoftsecbook>.
...
+implementation of low-level software. On the other hand, programs written in
+memory safe languages can themselves be vulnerable to memory errors as a result
+of bugs in their implementation. Can we fix the problem by switching to memory
```
kbeyls, line 145: "their implementation". When reading this I wondered if readers who are not compiler writers will very easily grasp what is meant by "implementation" here. If they stop to think about it, they for sure will understand. I wonder if using a few more words would help them to not have to stop and think about it at all, e.g. "... can themselves be vulnerable to memory errors as a result of bugs in how they are implemented."
I'm really not sure here, I just thought I'd share a thought I had here.

book.md
```diff
@@ -109,7 +109,73 @@ feedback to be received through <https://github.com/llsoftsec/llsoftsecbook>.
...
+safe languages for compiler and runtime implementation? Even if that were as
```
kbeyls, line 146: maybe "for THE compiler and runtime implementation TOO?"?

book.md
```diff
@@ -109,7 +109,73 @@ feedback to be received through <https://github.com/llsoftsec/llsoftsecbook>.
...
+simple as it sounds, unfortunately there are types of programming errors that
+these languages cannot protect against.
```
kbeyls, line 148: Maybe it would be good to be a little bit more explicit here too to convey the message more explicitly? Maybe something like "... that these languages cannot protect against. For example, a logical error in the implementation of a compiler or run-time of a memory safe language can lead to a memory access error not being detected.".
If that last sentence would be added, it would be great if we could refer to an example of that. I'm afraid I don't have any examples, but maybe you know of some?

book.md
```diff
@@ -109,7 +109,73 @@ feedback to be received through <https://github.com/llsoftsec/llsoftsecbook>.
...
+
+Given the rich history of memory vulnerabilities and mitigations and the active
+developments in this area, compiler developers are likely to encounter some of
+these issues over the course of their careers. This chapter aims to serve as an
+introduction to this area. We start with a discussion of exploitation
+primitives, which can be useful when discussing threat models. We then continue
```
kbeyls, line 154: We probably should have a \todo to also discuss threat models somewhere in the book too?
It's something orthogonal to specific vulnerabilities, so maybe it should go into a chapter of its own that we currently do not have? Or maybe just as a section in the introductory chapter?

book.md
```diff
@@ -109,7 +109,73 @@ feedback to be received through <https://github.com/llsoftsec/llsoftsecbook>.
...
+with a more detailed discussion of the various types of vulnerabilities, along
+with their mitigations, presented in a rough chronological order of their
+appearance (and therefore complexity).
```
kbeyls, line 157: "(and therefore complexity):" I seem to remember a style guideline to try and not put words in parentheses.
In this case, I think the "(and" can be replaced by “,and”, without the closing “)”.
I probably should look up what CMoS says, if anything, about use of parentheses, to check if my personal recollection of a random style guideline is actually “best practice”

book.md
g-kouv: I didn't find specific guidance in CMoS, but there are uses of "on the one hand", so I'll use this version too.

book.md
g-kouv: :)

book.md
g-kouv: It seems that Wikipedia has a whole article on how to cite it (and why you shouldn't...), and even an example BibTex entry: https://en.wikipedia.org/wiki/Wikipedia:Citing_Wikipedia#BibTeX_entry

I'll give this a go, but feel free to leave another comment if the new version I'll upload doesn't make sense.

kbeyls: Oh, I wouldn't cite, I would just make the text "Morris worm" a hyperlink to the Morris worm article on wikipedia, for readers that would like to have a starting point to explore more details about it.

I.e. something like [Morris worm](https://en.wikipedia.org/wiki/Morris_worm) in the markdown source code...

book.md
g-kouv: Yes, I intend to refer to such examples in the JIT section. For now, I'll add the text you suggested, and mention that we will see such examples in the JIT section (with a reference to it).

book.md
g-kouv: I tried the citation and it looks like overkill, to be honest, so I'll revert back to the simple hyperlink as you originally suggested.

book.md
```diff
@@ -109,7 +109,73 @@ feedback to be received through <https://github.com/llsoftsec/llsoftsecbook>.
  
 # Memory vulnerability based attacks and mitigations
 
-\missingcontent{Write chapter on memory vulnerabilities and mitigation.}
+## Introduction
```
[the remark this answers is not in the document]
g-kouv: I went with the word "background" as not all the content is about history.

PR #56, opened 2021-10-15
[proposal] Remove LaTeX todos and use task-lists instead.

fpetrogalli:
I would like to propose the removal of the use of the `todonotes` package in LaTeX in favor of the task lists that GitHub markdown supports natively: https://docs.github.com/en/issues/tracking-your-work-with-issues/about-task-lists

I see two main advantages in doing so:

1. The tasks are rendered in the actual [pages showed on github](https://github.com/fpetrogalli/llsoftsecbook/blob/task-lists/book.md).
2. These task lists integrate natively with the issues system of github. For details see https://docs.github.com/en/issues/tracking-your-work-with-issues/about-task-lists#about-issue-task-lists.

This will make issues and todos more visible and traceable in github directly. In particular, tasks linked to an issue or pr will be checked as done automatically when the issue or the pr are closed.

The tasks render pretty well  in the PDF too, with an empty checkbox.

Using tasks also make the CI check the also the links that were previously skipped because mentioned in LaTeX todos via`\href`.

The only thing that will be missing is the list of todos in the PDF, but why would people want to see what is left to do by opening the PDF when they can simply click on the issue tab or check the file `book.md` on the web interface of ithub?

Anyway - I might be missing some reasons why you are using `todonotes`. Please feel free to discard this proposal if you think it does not suite the needs of the project!

### DIFF
--- a/book.md
+++ b/book.md
@@ -14,18 +14,6 @@ header-includes:
   ```{=latex}
   \usepackage{makeidx}
   \makeindex
-  \newcounter{TodoCounter}
-  \usepackage[backgroundcolor=white,linecolor=black]{todonotes}
-  \let\oldtodo\todo
-  \usepackage{bclogo}%  \bcpanchant
-  \renewcommand{\todo}[1]{
-    \stepcounter{TodoCounter}
-    \oldtodo[caption={\arabic{TodoCounter}. #1}]{\bcpanchant #1}
-  }
-  \newcommand{\missingcontent}[1]{
-    \stepcounter{TodoCounter}
-    \oldtodo[inline,caption={\arabic{TodoCounter}. #1}]{\bcpanchant \textit{#1}}
-  }
   ```
 ...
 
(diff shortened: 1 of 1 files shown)


kbeyls:
I really like how this proposal would enable a better link with todos written in the book and github issues and pull requests.

I do think the task lists are rendered less nicely than the todos are currently rendered in the PDF with the latex package. I think they are more confusing to readers - i.e. it's less clear they are todo notes rather than actual book content.
I was also thinking to have todos nicely formatted in the HTML output, taking inspiration from https://kennethfriedman.org/thoughts/2019/marginal-notes/ .

I also think it would be good to have versions of the PDF and HTML book with and without nicely rendered TODOs.

I think we can have all advantages of all the above by having a custom pandoc lua script that would translate the markdown tasklist to the appropriate format in the LaTeX and HTML output.

fpetrogalli:
Abandoning this as I misunderstood the way task list are handled in github.

PR #70, opened 2021-12-10
Add a few words on copyright and license to contributing.md

kbeyls:


contributing.md
```diff
@@ -35,6 +35,20 @@ Please send a GitHub Pull Request (read more about
...
 [style and coding conventions](#style-and-coding-conventions).
 
+### Copyright and license
+
+The copyright on contributions is not transferred. In other words, contributors
+retain copyright on their contributions.
+
+If you're making a contribution for the first time, please check if the
+copyright owner is listed in the header under the "copyright:" tag at
+https://raw.githubusercontent.com/llsoftsec/llsoftsecbook/main/book.md. If not,
```
JLouisKaplan-Arm, line 45: Might want to link to the 'source blob' rather than the raw file if you think the surrounding info (line number, recent commits etc.) is useful:

https://github.com/llsoftsec/llsoftsecbook/blob/main/book.md?plain=1

Could use this to have a linked line reference, though maybe this is too fragile in case the copyright line changes in the future:

https://github.com/llsoftsec/llsoftsecbook/blob/main/book.md?plain=1#L3

JLouisKaplan-Arm (changes requested):
Looks good to me. Very clear. Just one small suggestion, but I don't feel strongly about it.

Thanks!

contributing.md
kbeyls: Nice, I didn't know about the "?plain=1" option. I'll use that, but not the "#L3" version as indeed the line might change in the future.

kbeyls:
@all-contributors please add @JLouisKaplan-Arm for review

kbeyls:
@all-contributors please add @JLouisKaplan-Arm for review

Issue #32, opened 2021-10-04
Make HTML version of book mobile device friendly.


g-kouv:
The HTML version of the book here: https://llsoftsec.github.io/llsoftsecbook/ should be made easier to read on mobile devices.

JLouisKaplan-Arm:
I haven't much CSS knowledge, but something like this might be enough as a temporary stop-gap?
https://github.com/JLouisKaplan-Arm/llsoftsecbook/blob/gh-pages/responsive_small.css

the effects of which can be seen (with width less than 768px) at:
https://jlouiskaplan-arm.github.io/llsoftsecbook/

The text justification needs a little further tweaking.

kbeyls:
I like it :)
When I tried it on my mobile phone, the look-and-feel was as I expected, i.e. making the book readable.

@g-kouv , I wonder what you think?

JLouisKaplan-Arm:
_small update:_ returned headings to being left aligned, and only justified <p> elements, which looks a little neater to my eyes (no big gaps in words of headings).

kbeyls:
I've been looking for examples in the past few days of style sheets used in other online books written using pandoc.
I only found 2 categories:
1. https://jez.io/tufte-pandoc-css/ (source code at https://github.com/jez/tufte-pandoc-css)
2. The books written using bookdown, see examples at https://bookdown.org/.

Both categories seem to have style sheets that are "responsive", i.e. adapt well to different screen sizes, such as mobile or desktop.

For tufte-pandoc-css, I particularly like:
* Footnotes appear as side notes in the html rendering.
* The example at https://jez.io/tufte-pandoc-css/ doesn't show it, but it is possible to have the table of contents as side notes.
* I just personally like the overall look-and-feel.

* A disadvantage is that it seems that the license the CSS file is under may be incompatible with the CC-BY license used for llsoftsecbook.

For the examples at https://bookdown.org/, I liked:
* It makes use of bootstrap, which should make it easier to adapt/tweak it further.

I like the overall look-and-feel less though - it seems there are more buttons etc - making the books feel more like a website than a "book". Of course, this is very subjective.

Overall, I think that maybe the best thing to do is to just bite the bullet and start writing our own CSS.

kbeyls:
I think we can close this issue now.
The HTML version of the book is now as good (or as bad) as the large-screen version.
We can drive further improvements to the HTML output (both for mobile and large-screen versions) through #130

PR #79, opened 2022-01-17
Make small/mobile screen view more readable

JLouisKaplan-Arm:
This commit addresses issue #32 by adding some basic 'responsive design' CSS to the default.css file such that the page layout is minimally adjusted when the viewport is less than 720px in width.

The effects of this change can be seen at:
https://jlouiskaplan-arm.github.io/llsoftsecbook/ 

(Note the changes only take effect with a viewport width of less than 720px.)

The main changes made for small screens are moving the contents above the paragraph elements and justifying paragraph text.

The advantage of the 'responsive design' approach is that changes made for small screens have no effect on the large screen styling.

### DIFF
--- a/default.css
+++ b/default.css
@@ -17,3 +17,26 @@ body {
 body {
   font-family: sans-serif;
 }
+
+/* For smaller screens: */
+@media only screen and (max-width : 768px) {
+/* Put table-of-contents above paragraph elements. */
+#TOC {
+  position: relative;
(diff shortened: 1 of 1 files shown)


g-kouv (approved):
Looks good to me, thanks!

kbeyls:
LGTM, thank you!

kbeyls (approved):

PR #80, opened 2022-02-07
Add section on exploitation primitives

g-kouv:
An introduction to read/write primitives and how they can be used. PTAL!

### DIFF
--- a/book.bib
+++ b/book.bib
@@ -87,7 +87,7 @@ Please keep the comma after the author's name
   author = {Matt Miller},
   title  = {Modeling the exploitation and mitigation of memory safety vulnerabilities},
   url    = {https://github.com/Microsoft/MSRC-Security-Research/blob/master/presentations/2012_10_Breakpoint/BreakPoint2012_Miller_Modeling_the_exploitation_and_mitigation_of_memory_safety_vulnerabilities.pdf},
-  year   = {2012},
+  howpublished = {\href{https://2012.ruxconbreakpoint.com/}{Breakpoint 2012}},
 }
 
 @inproceedings{Hu2016,
--- a/book.bib
+++ b/book.bib
@@ -83,3 +83,52 @@ Please keep the comma after the author's name
   journal = {WIRED},
 }
 
+@Electronic{Miller2012,
+  author = {Matt Miller},
+  title  = {Modeling the exploitation and mitigation of memory safety vulnerabilities},
+  url    = {https://github.com/Microsoft/MSRC-Security-Research/blob/master/presentations/2012_10_Breakpoint/BreakPoint2012_Miller_Modeling_the_exploitation_and_mitigation_of_memory_safety_vulnerabilities.pdf},
+  year   = {2012},
+}
+
+@inproceedings{Hu2016,
+  author    = {Hu, Hong and Shinde, Shweta and Adrian, Sendroiu and Chua, Zheng Leong and Saxena, Prateek and Liang, Zhenkai},
+  booktitle = {2016 IEEE Symposium on Security and Privacy (SP)},
+  title     = {Data-Oriented Programming: On the Expressiveness of Non-control Data Attacks},
+  year      = {2016},
+  volume    = {},
+  number    = {},
+  pages     = {969-986},
+  doi       = {10.1109/SP.2016.62}
+}
+
+@article{Dullien2020,
+  author    = {Dullien, Thomas},
+  journal   = {IEEE Transactions on Emerging Topics in Computing},
+  title     = {Weird Machines, Exploitability, and Provable Unexploitability},
+  year      = {2020},
+  volume    = {8},
+  number    = {2},
+  pages     = {391-403},
+  doi       = {10.1109/TETC.2017.2785300}
+}
+
+@Electronic{Beer2020,
+  author = {Ian Beer},
+  title  = {An iOS zero-click radio proximity exploit odyssey},
+  url    = {https://googleprojectzero.blogspot.com/2020/12/an-ios-zero-click-radio-proximity.html},
+  year   = {2020},
+}
+
+@Electronic{Groß2020,
+  author = {Samuel Groß},
+  title  = {JITSploitation I: A JIT Bug},
+  url    = {https://googleprojectzero.blogspot.com/2020/09/jitsploitation-one.html},
+  year   = {2020},
+}
+
+@Electronic{Beer2021,
+  author = {Ian Beer and Samuel Groß},
+  title  = {A deep dive into an NSO zero-click iMessage exploit: Remote Code Execution},
+  url    = {https://googleprojectzero.blogspot.com/2021/12/a-deep-dive-into-nso-zero-click.html},
+  year   = {2021},
+}
--- a/book.md
+++ b/book.md
@@ -177,7 +177,146 @@ vulnerabilities, along with their mitigations, presented in a rough
 chronological order of their appearance, and, therefore, complexity.
 
 ## Exploitation primitives
-\missingcontent{Discuss exploitation primitives}
+
+Newcomers to the area of software security may find themselves lost in a large
+number of blog posts and other publications describing specific memory
+vulnerabilities and how to exploit them. Two very common, yet unfamiliar to a
+newcomer, terms that appear in such publications are _read primitive_ and
+_write primitive_. In order to understand memory vulnerabilities and be able to
+design effective mitigations, it's important to understand what these terms
+mean, how these primitives could be achieved by an attacker, and how they can
+be used.
+
+An _exploit primitive_ is a mechanism that allows an attacker to reliably
+perform a specific operation in the memory space of the victim program, through
+an input interaction with the victim program.
+
+A _write primitive_ gives the attacker write access to the victim's memory space.
+The value written and the address written to may be controlled by the attacker
+to various degrees. The primitive, for example, may allow:
+
+* writing a fixed value to an attacker-controlled address
+* writing to an address consisting of a fixed base and an attacker-controlled
+  offset limited to a specific range (e.g. a 32-bit offset)
+* writing to an attacker-controlled base address with a fixed offset
+
+Slide 11 of [@Miller2012] describes more properties of primitives that allow a
+more detailed classification.
+
+The most powerful version of a write primitive is an _arbitrary write_
+primitive, where both the address and the value are fully controlled by the
+attacker.
+
+A _read primitive_, respectively, gives the attacker read access to the
+victim's memory space. As for the write primitive, the address of the memory
+location accessed will be controlled by the attacker to some degree. One of the
+attacker's early goals will often be to achieve an _arbitrary read_ primitive,
+in which the address is fully controlled by the attacker.
+
+The effects of a write primitive are perhaps easier to understand, as it
+has obvious side-effects: a value is written to the victim program's memory.
+But how does an attacker get to observe the result of a read primitive?
+
+This basically depends on whether the attack is interactive or non-interactive,
+terminology which is taken from [@Hu2016].
+
+* In an _interactive attack_, the attacker gives malicious input to the victim
+  program, observes the result, and prepares the next input. Observing the
+  results is done through looking at the victim program's output, for example
+  looking at a network packet transmitted by the victim. For an attacker to
+  be able to interact with the victim through these primitives, the victim
+  process must stay alive between the handling of the malicious inputs, i.e.
+  it must not be restarted. Note that although the crafting of malicious
+  input could in theory be done by hand, it's usually done via an external
+  program that drives the exploit. An example of this type of attack can be
+  seen in [@Beer2020], which describes a zero-click radio proximity
+  exploit.
+* In a _non-interactive (one-shot) attack_, the attacker provides all malicious
+  input to the victim program at once. The malicious input triggers multiple
+  primitives one after the other, and the primitives are able to observe the
+  effects of the preceding operations through the victim program's state.
+  The input could be, for example, in the form of a JavaScript program
+  [@Groß2020], or a PDF file pretending to be a GIF [@Beer2021].
+
+How does an attacker achieve these kinds of primitives in the first place?
+The details for this vary, and in some cases it takes a combination of many
+techniques, some of which are out of scope for this book. But we will be
+seeing a few of them in this chapter, for example a stack overflow
+resulting in a (restricted) write primitive when the input size exceeds what
+the program expected.
+
+How are primitives chained to perform multiple reads/writes?
+In the case of an interactive attack, preparing and sending input to the victim
+program and parsing the output of the victim program are usually done in an
+external program that drives the exploit. This could be, for example, a C
+program communicating with the victim over the Internet. In this case, the
+primitives are abstracted into C functions, which prepare and send packets to
+the victim, and parse the victim's responses. These functions can be easily
+chained together, interleaved with arbitrary computations, all written in C, to
+form the exploit.
+
+It's interesting to note that while the read/write primitives consist of
+carefully constructed inputs to the victim program, the attacker can view these
+inputs as *instructions* to the victim program). This duality is
+explored in [@Dullien2020].
+
+In the case of a non-interactive attack, all computation happens within the
+victim program. The duality of input data and code is even more obvious in this
+case, as the malicious input to the victim can be viewed as the exploit code.
+There are cases for which the input is obviously interpreted as code by the
+victim application as well, as in the case of a JavaScript program given as
+input to a JavaScript engine. In this case, the read/write primitives would
+be written as JavaScript functions, which when called have the unintended
+side-effect of accessing arbitrary memory that a JavaScript program is not
+supposed to have access to.  The primitives can be chained together with
+arbitrary computations, also expressed in JavaScript.
+
+There are, however, cases, where the correspondence between data and code isn't
+as obvious. For example, in [@Beer2021], the malicious input consists of a PDF
+file, masquerading as a GIF. Due to an integer overflow bug in the PDF decoder,
+the malicious input leads to an unbounded buffer access.  In the case of
+JavaScript engine exploitation, the attacker would normally be able to use
+JavaScript operations and perform arbitrary computations, making exploitation
+more straightforward. In this case, there are no scripting capabilities
+officially supported. The attackers, however, take advantage of the compression
+format intricacies to implement a small computer architecture, in thousands of
+simple commands to the decoder.  In this way, they effectively _introduce_
+scripting capabilities and are able to express their exploit as a program to
+this architecture.
+
+We have described read/write primitives, and have discussed how an attacker
+might also perform arbitrary computations, either in an external program
+in the case of interactive attacks, or by using scripting capabilities
+(whether originally supported or introduced by the attacker) in
+non-interactive attacks. Assuming an attacker has gained these capabilities,
+how can they use them to achieve their goals?
+
+The ultimate goal of an attacker may vary: it may be, among other things,
+getting access to a system, leaking sensitive information or bringing down a
+service. Frequently, a first step towards these wider goals is arbitrary code
+execution within the victim process. We have already mentioned that the
+attacker will typically have arbitrary computation capabilities at this point,
+but arbitrary code execution also involves things like calling arbitrary
+library functions, performing system calls, and so on.
+
+Some examples of how the attacker may use the obtained primitives:
+
+* Leak information, such as pointers to specific data structures or code,
+  or the stack pointer.
+* Overwrite the stack contents, e.g. to perform a [ROP attack](#code-reuse-attacks).
+* Overwrite non-control data, e.g. authorization state. Sometimes this
+  step is sufficient to achieve the attacker's goal, bypassing the need for
+  arbitrary code execution.
+
+Once arbitrary execution is achieved, the attacker may need to exploit
+additional vulnerabilities in order to escape a process sandbox, escalate
+privilege, etc. Such vulnerability chaining is common, but for the
+purposes of this chapter we will focus on:
+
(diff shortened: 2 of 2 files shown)


book.md
```diff
@@ -177,7 +177,146 @@ vulnerabilities, along with their mitigations, presented in a rough
 chronological order of their appearance, and, therefore, complexity.
 
 ## Exploitation primitives
-\missingcontent{Discuss exploitation primitives}
+
+Newcomers to the area of software security may find themselves lost in a large
+number of blog posts and other publications describing specific memory
```
sam-ellis, line 182: Possible simplification of "a large number of" to "many".

book.md
```diff
@@ -177,7 +177,146 @@ vulnerabilities, along with their mitigations, presented in a rough
...
+non-interactive attacks. Assuming an attacker has gained these capabilities,
+how can they use them to achieve their goals?
+
+The ultimate goal of an attacker may vary: it may be, among other things,
+getting access to a system, leaking sensitive information or bringing down a
+service. Frequently, a first step towards these wider goals is arbitrary code
+execution within the victim process. We have already mentioned that the
+attacker will typically have arbitrary computation capabilities at this point,
+but arbitrary code execution also involves things like calling arbitrary
+library functions, performing system calls, and so on.
```
sam-ellis, line 300: Having both "things like" and "and so on" is redundant, and could possibly be simplified to just one of these.

book.md
```diff
@@ -177,7 +177,146 @@ vulnerabilities, along with their mitigations, presented in a rough
...
+perform a specific operation in the memory space of the victim program, through
+an input interaction with the victim program.
+
+A _write primitive_ gives the attacker write access to the victim's memory space.
+The value written and the address written to may be controlled by the attacker
+to various degrees. The primitive, for example, may allow:
+
+* writing a fixed value to an attacker-controlled address
+* writing to an address consisting of a fixed base and an attacker-controlled
+  offset limited to a specific range (e.g. a 32-bit offset)
```
sam-ellis, line 200: Is it important that the offset has a limited range? Could this be simplified to be just "an attacker-controlled offset"?

book.md
g-kouv: Good point! I'll make the change.

book.md
g-kouv: Indeed, changing this.

book.md
sam-ellis: While it matters a lot in a specific case whether the offset is signed or not and it's size and other properties of it, I wonder if it is too much detail here in a place that looks more like a summary. I was thinking it might be sufficient to say a fixed base + attacker-controlled offset. On the other hand, if this will be the only mention of this in the book, then perhaps more detail is required such as explicitly mentioning widening and sign-extension as variations.

book.md
```diff
@@ -177,7 +177,146 @@ vulnerabilities, along with their mitigations, presented in a rough
...
+  [@Groß2020], or a PDF file pretending to be a GIF [@Beer2021].
+
+How does an attacker achieve these kinds of primitives in the first place?
+The details for this vary, and in some cases it takes a combination of many
+techniques, some of which are out of scope for this book. But we will be
+seeing a few of them in this chapter, for example a stack overflow
+resulting in a (restricted) write primitive when the input size exceeds what
+the program expected.
+
+How are primitives chained to perform multiple reads/writes?
```
kbeyls, line 248: I think this is the first place in the text that refers to chaining primitives?
If so, I think that maybe a paragraph is needed above this to first explain what chaining primitives is and why it's needed. Then it's easier for the reader to then read about the how.

book.md
```diff
@@ -177,7 +177,146 @@ vulnerabilities, along with their mitigations, presented in a rough
...
+In the case of an interactive attack, preparing and sending input to the victim
+program and parsing the output of the victim program are usually done in an
+external program that drives the exploit. This could be, for example, a C
```
kbeyls, line 251: I'm wondering in how far it matters that the communicating program is written in C?
I think that in the few exploits I've seen described making use of this technique, the program was written in python.

book.md
```diff
@@ -177,7 +177,146 @@ vulnerabilities, along with their mitigations, presented in a rough
...
+program communicating with the victim over the Internet. In this case, the
+primitives are abstracted into C functions, which prepare and send packets to
+the victim, and parse the victim's responses. These functions can be easily
+chained together, interleaved with arbitrary computations, all written in C, to
+form the exploit.
```
kbeyls, line 256: It would be good to have a reference here where the reader can get a more detailed explanation, for example to 
a tutorial explaining this in more detail.

book.md
```diff
@@ -177,7 +177,146 @@ vulnerabilities, along with their mitigations, presented in a rough
...
+
+It's interesting to note that while the read/write primitives consist of
```
kbeyls, line 258: A fundamentally new concept gets introduced here. Maybe it'd be best to have a new subsection heading here, e.g. "Weird machines"?

book.md
```diff
@@ -177,7 +177,146 @@ vulnerabilities, along with their mitigations, presented in a rough
...
+carefully constructed inputs to the victim program, the attacker can view these
+inputs as *instructions* to the victim program). This duality is
```
kbeyls, line 260: Rather than "instructions to the victim program", I'd probably write "intructions to an interpreter that the victim program happens to un-intentionally implement". Not sure if that would be easier to understand for most readers though...

book.md
```diff
@@ -177,7 +177,146 @@ vulnerabilities, along with their mitigations, presented in a rough
...
+input to a JavaScript engine. In this case, the read/write primitives would
+be written as JavaScript functions, which when called have the unintended
+side-effect of accessing arbitrary memory that a JavaScript program is not
+supposed to have access to.  The primitives can be chained together with
+arbitrary computations, also expressed in JavaScript.
+
+There are, however, cases, where the correspondence between data and code isn't
+as obvious. For example, in [@Beer2021], the malicious input consists of a PDF
+file, masquerading as a GIF. Due to an integer overflow bug in the PDF decoder,
+the malicious input leads to an unbounded buffer access.  In the case of
```
kbeyls, line 277: "to an unbounded buffer access, which is an arbitrary read primitive"?

book.md
```diff
@@ -177,7 +177,146 @@ vulnerabilities, along with their mitigations, presented in a rough
...
+JavaScript operations and perform arbitrary computations, making exploitation
+more straightforward. In this case, there are no scripting capabilities
+officially supported. The attackers, however, take advantage of the compression
+format intricacies to implement a small computer architecture, in thousands of
+simple commands to the decoder.  In this way, they effectively _introduce_
+scripting capabilities and are able to express their exploit as a program to
+this architecture.
+
+We have described read/write primitives, and have discussed how an attacker
+might also perform arbitrary computations, either in an external program
```
kbeyls, line 288: I'd split the sentence at the comma between "computations" and "either".

kbeyls (approved):
This is an amazing description. Thank you very much @g-kouv !
I have added lots of detailed comments and questions, but feel free to disagree with them.
I also think that most of them could be addressed in follow-up patches; let's not make the perfect the enemy of the good!
If you'd decide to do some of them later, please feel free to note the ideas for improvements in the source using \todo{}s.

book.md
```diff
@@ -177,7 +177,146 @@ vulnerabilities, along with their mitigations, presented in a rough
...
+case, as the malicious input to the victim can be viewed as the exploit code.
+There are cases for which the input is obviously interpreted as code by the
+victim application as well, as in the case of a JavaScript program given as
+input to a JavaScript engine. In this case, the read/write primitives would
+be written as JavaScript functions, which when called have the unintended
+side-effect of accessing arbitrary memory that a JavaScript program is not
+supposed to have access to.  The primitives can be chained together with
+arbitrary computations, also expressed in JavaScript.
+
+There are, however, cases, where the correspondence between data and code isn't
```
JLouisKaplan-Arm, line 274: extra comma after cases?

book.md
```diff
@@ -177,7 +177,146 @@ vulnerabilities, along with their mitigations, presented in a rough
...
+
+We have described read/write primitives, and have discussed how an attacker
+might also perform arbitrary computations, either in an external program
+in the case of interactive attacks, or by using scripting capabilities
+(whether originally supported or introduced by the attacker) in
+non-interactive attacks. Assuming an attacker has gained these capabilities,
+how can they use them to achieve their goals?
+
+The ultimate goal of an attacker may vary: it may be, among other things,
+getting access to a system, leaking sensitive information or bringing down a
```
JLouisKaplan-Arm, line 295: could leaking sensitive information or bringing down a service be considered subtypes of getting access to a system, I think might have the wrong idea here?

JLouisKaplan-Arm (commented):
From my beginner perspective, I found this a really interesting and clear read. I hope it's OK that I added a few comments. They are mostly questions for my own curiousity - feel free to ignore!

book.md
```diff
@@ -177,7 +177,146 @@ vulnerabilities, along with their mitigations, presented in a rough
...
+Newcomers to the area of software security may find themselves lost in a large
+number of blog posts and other publications describing specific memory
+vulnerabilities and how to exploit them. Two very common, yet unfamiliar to a
+newcomer, terms that appear in such publications are _read primitive_ and
+_write primitive_. In order to understand memory vulnerabilities and be able to
+design effective mitigations, it's important to understand what these terms
+mean, how these primitives could be achieved by an attacker, and how they can
+be used.
+
+An _exploit primitive_ is a mechanism that allows an attacker to reliably
```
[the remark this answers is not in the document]
g-kouv: Removing.

book.md
```diff
@@ -177,7 +177,146 @@ vulnerabilities, along with their mitigations, presented in a rough
...
+perform a specific operation in the memory space of the victim program, through
+an input interaction with the victim program.
```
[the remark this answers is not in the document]
g-kouv: I'm changing this, splitting it into a separate sentence as suggested below. I won't add the word "typically", though, to keep this simpler, since for this chapter this is the only kind of interaction we consider. (Side channels, for example, are described elsewhere).

book.bib
```diff
@@ -83,3 +83,52 @@ @article{Greenberg2019
   journal = {WIRED},
 }
 
+@Electronic{Miller2012,
+  author = {Matt Miller},
+  title  = {Modeling the exploitation and mitigation of memory safety vulnerabilities},
+  url    = {https://github.com/Microsoft/MSRC-Security-Research/blob/master/presentations/2012_10_Breakpoint/BreakPoint2012_Miller_Modeling_the_exploitation_and_mitigation_of_memory_safety_vulnerabilities.pdf},
+  year   = {2012},
+}
```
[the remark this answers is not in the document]
g-kouv: This would be useful. I can't find an appropriate field, though - do you know what I could use for this?

book.md
```diff
@@ -177,7 +177,146 @@ vulnerabilities, along with their mitigations, presented in a rough
...
+The value written and the address written to may be controlled by the attacker
+to various degrees. The primitive, for example, may allow:
+
+* writing a fixed value to an attacker-controlled address
+* writing to an address consisting of a fixed base and an attacker-controlled
+  offset limited to a specific range (e.g. a 32-bit offset)
+* writing to an attacker-controlled base address with a fixed offset
+
+Slide 11 of [@Miller2012] describes more properties of primitives that allow a
+more detailed classification.
```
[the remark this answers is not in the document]
g-kouv: Rephrased along those lines.

book.bib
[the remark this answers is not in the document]
kbeyls: Maybe if you changed the type of entry from "@Electronic" to "@Inproceedings", which seems to be the typical entry type used in bibtex for conference proceedings, then maybe there would be an obvious field to put this in?

book.md
```diff
@@ -177,7 +177,146 @@ vulnerabilities, along with their mitigations, presented in a rough
...
+
+The most powerful version of a write primitive is an _arbitrary write_
+primitive, where both the address and the value are fully controlled by the
+attacker.
+
+A _read primitive_, respectively, gives the attacker read access to the
+victim's memory space. As for the write primitive, the address of the memory
```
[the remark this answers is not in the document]
g-kouv: "Just like.. " doesn't seem natural to me, so I've instead moved the phrase to the end of the sentence, which hopefully makes it less ambiguous.

book.md
```diff
@@ -177,7 +177,146 @@ vulnerabilities, along with their mitigations, presented in a rough
...
+location accessed will be controlled by the attacker to some degree. One of the
+attacker's early goals will often be to achieve an _arbitrary read_ primitive,
+in which the address is fully controlled by the attacker.
+
+The effects of a write primitive are perhaps easier to understand, as it
+has obvious side-effects: a value is written to the victim program's memory.
+But how does an attacker get to observe the result of a read primitive?
+
+This basically depends on whether the attack is interactive or non-interactive,
+terminology which is taken from [@Hu2016].
```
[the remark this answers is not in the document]
g-kouv: I'm not sure this is the best reference for this terminology, since it's not the paper's main topic, which is why I felt the need to add the wordy explanation here. It's better than no reference, though. Removing the unnecessary phrase.

book.md
```diff
@@ -177,7 +177,146 @@ vulnerabilities, along with their mitigations, presented in a rough
...
+  seen in [@Beer2020], which describes a zero-click radio proximity
+  exploit.
+* In a _non-interactive (one-shot) attack_, the attacker provides all malicious
+  input to the victim program at once. The malicious input triggers multiple
+  primitives one after the other, and the primitives are able to observe the
+  effects of the preceding operations through the victim program's state.
+  The input could be, for example, in the form of a JavaScript program
+  [@Groß2020], or a PDF file pretending to be a GIF [@Beer2021].
+
+How does an attacker achieve these kinds of primitives in the first place?
```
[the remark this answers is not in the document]
g-kouv: I've changed the couple of instances I could find to "obtain", since "achieve" seemed confusing, but I don't think it makes a huge difference. Most of the time these terms are used informally, for example the word "primitive" is often skipped altogether. "Achieve arbitrary read/write" is not uncommon. You'll also find "create", "construct", or, even more commonly, simply "get". I don't think there's a lot of consensus on terminology.

Perhaps that's something we could have a whole different *chapter* on: expectations when reading exploit/CTF writeups and related writings. These are very relevant for memory vulnerability exploits, but they also tend to be a lot more informal than a lot of the literature that a compiler engineer might be used to, with terms that don't necessarily have textbook definitions. It's hard to use the terminology exactly as in the kinds of writeups where it's found (a little more casually), while at the same time catering to the compiler engineer (more pedantic than average) audience.

book.md
g-kouv: It doesn't matter which language is used, it's just an example. I've picked one language and stuck with it, so I can say things like "abstracted into C functions" later in this paragraph, which were wordy and confusing without a reference to a specific language.

book.md
g-kouv: I'm not sure it's clearer. In my experience (based on people's reactions to my longer sentences), shorter sentences seem to work better. Instead of making this sentence longer, I'll elaborate in a separate sentence, which should be a fair compromise between the two.

book.md
```diff
@@ -177,7 +177,146 @@ vulnerabilities, along with their mitigations, presented in a rough
...
 ## Exploitation primitives
-\missingcontent{Discuss exploitation primitives}
+
+Newcomers to the area of software security may find themselves lost in a large
+number of blog posts and other publications describing specific memory
+vulnerabilities and how to exploit them. Two very common, yet unfamiliar to a
+newcomer, terms that appear in such publications are _read primitive_ and
+_write primitive_. In order to understand memory vulnerabilities and be able to
+design effective mitigations, it's important to understand what these terms
+mean, how these primitives could be achieved by an attacker, and how they can
```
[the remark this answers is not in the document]
g-kouv: Obtaining the primitive could indeed be a goal in itself in some cases, for example if the attacker's goal is to leak memory contents that become accessible with that primitive. In other cases, it will be a step towards a larger goal, so a sub-goal.

The words achieve and attain seem to be synonyms (according to online dictionaries). It seems to me that 'attain' is less common and perhaps a little less friendly to non-native speakers. Since both you and Kristof (there's a more detailed discussion in his review) objected to 'achieve', I'm going with 'obtain'!

book.md
```diff
@@ -177,7 +177,146 @@ vulnerabilities, along with their mitigations, presented in a rough
...
+be used.
+
+An _exploit primitive_ is a mechanism that allows an attacker to reliably
+perform a specific operation in the memory space of the victim program, through
```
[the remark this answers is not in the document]
g-kouv: What would performing an operation outside of the memory space of the victim program look like? The way I meant the phrase, performing addition on a calculator, or running an arbitrary Python program on your own machine would fit that description, and those indeed would not be exploit primitives. But I'm not really trying to define very strictly what *isn't* useful for exploitation, I'm just trying to explain what the word "primitive" means when we talk about read/write primitives, since those terms are used a lot, and the first impression of what they are might be inaccurate (is it a "gadget"? is it some other code?). We're focussing on memory-based vulnerabilities specifically in this chapter, so restricting ourselves to talking about the memory space of the victim program is useful!

book.md
g-kouv: I was actually trying to avoid referring to "weird machines" explicitly (while making the reference available for the interested reader).

Subsection headings may be useful regardless, though, but it's be something I'd rather leave as a future improvement. But of course, as always, additional patches are welcome, if you feel inclined to do this asap!

book.md
```diff
@@ -177,7 +177,146 @@ vulnerabilities, along with their mitigations, presented in a rough
...
+an input interaction with the victim program.
+
+A _write primitive_ gives the attacker write access to the victim's memory space.
+The value written and the address written to may be controlled by the attacker
+to various degrees. The primitive, for example, may allow:
+
+* writing a fixed value to an attacker-controlled address
+* writing to an address consisting of a fixed base and an attacker-controlled
+  offset limited to a specific range (e.g. a 32-bit offset)
+* writing to an attacker-controlled base address with a fixed offset
```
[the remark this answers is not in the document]
g-kouv: This seems to be the relevant section: https://www.chicagomanualofstyle.org/book/ed17/part2/ch06/psec130.html

Picking some quotes that seem to apply:

> Closing punctuation is used only if items consist of complete sentences. 

> If none of the items in a bulleted list consist of complete sentences, however, each item can usually begin lowercase (except for proper nouns). For bulleted lists whose items require more prominence, capitalization may instead be preferred (as throughout this manual); choose one approach and follow it consistently.

My list doesn't contain complete sentences, so I think it's OK as it is. However, we can decide on a different (consistent) style, in which case it would be useful to document it.

book.md
```diff
@@ -177,7 +177,146 @@ vulnerabilities, along with their mitigations, presented in a rough
...
+
+The most powerful version of a write primitive is an _arbitrary write_
+primitive, where both the address and the value are fully controlled by the
+attacker.
+
+A _read primitive_, respectively, gives the attacker read access to the
+victim's memory space. As for the write primitive, the address of the memory
+location accessed will be controlled by the attacker to some degree. One of the
+attacker's early goals will often be to achieve an _arbitrary read_ primitive,
+in which the address is fully controlled by the attacker.
```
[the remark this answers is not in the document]
g-kouv: Since I'm talking more about attacker goals later, I've removed the reference here, instead stating that an arbitrary write primitive is a particularly useful primitive, in order to introduce it.

book.md
```diff
@@ -177,7 +177,146 @@ vulnerabilities, along with their mitigations, presented in a rough
...
+
+This basically depends on whether the attack is interactive or non-interactive,
+terminology which is taken from [@Hu2016].
+
+* In an _interactive attack_, the attacker gives malicious input to the victim
+  program, observes the result, and prepares the next input. Observing the
+  results is done through looking at the victim program's output, for example
+  looking at a network packet transmitted by the victim. For an attacker to
+  be able to interact with the victim through these primitives, the victim
+  process must stay alive between the handling of the malicious inputs, i.e.
```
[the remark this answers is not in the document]
g-kouv: Rephrasing along those lines.

book.md
```diff
@@ -177,7 +177,146 @@ vulnerabilities, along with their mitigations, presented in a rough
...
+  it must not be restarted. Note that although the crafting of malicious
+  input could in theory be done by hand, it's usually done via an external
+  program that drives the exploit. An example of this type of attack can be
+  seen in [@Beer2020], which describes a zero-click radio proximity
+  exploit.
```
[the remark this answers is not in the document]
g-kouv: I use the word "describe" immediately after, so I was looking for an alternative. I'll rephrase to the simpler "You can see...", there's no need for passive voice here.

book.md
g-kouv: I'll rephrase this part, including using the word "combine" instead of chain, at least in the first instance, to make this more self-evident.

g-kouv:
Thanks for the comments, @sam-ellis , @kbeyls and @JLouisKaplan-Arm!

I've just uploaded my updates (I tried to address most comments, or leave a TODO, but might have missed something since there were so many). Since Kristof has already approved, I intend to push the updated version tomorrow unless there's a strong objection.

book.md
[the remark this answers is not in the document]
JLouisKaplan-Arm: Sounds good to me!!

book.md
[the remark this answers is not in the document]
JLouisKaplan-Arm: Yep I see what you mean. Also, thinking about this again, I realise that I was tying myself in unnecessary knots here; a mechanism to attack a subsequent (not the initial attack target) program's memory space is a primitive with respect to that program. All looks good to me!

book.md
JLouisKaplan-Arm: No I think they're helpful. In hindsight, I was over-analyzing it. I tried for a while to come up with some alternative wordings and clarify my thoughts, but in the end the text that you have already covers it all. LGTM.

PR #83, opened 2022-02-11
Make headers clickable links to self

### LINKED ISSUE #64: Make headings and subheadings directly linkable
As the book gets bigger, it will likely become increasingly useful to be able to link directly to headings and subheadings (both externally, and crosslinking from book itself). Maybe Pandoc has a setting to enable this?

JLouisKaplan-Arm:
Fixes #64 

* Makes headers (h1, h2, h3) show
  a section symbol on hover/click
  which is a link to itself.

* Also some basic reorganization for
  neatness, planning ahead for potentially
  more complex frontend customisation.

Effects viewable at: https://jlouiskaplan-arm.github.io/llsoftsecbook/

(Note, make sure browser hasn't cached an old `.css` style file.)

On mobile // touch-screens: I tried to make the behaviour as inuitive as I could. One tap brings up the link icon, and only one header can be 'active' at a time.


kbeyls (approved):
This is amazing, thank you for implementing this!
I only noticed some seemingly accidental white space changes, otherwise LGTM.
I'm afraid I don't know enough lua to be absolutely sure exactly what the filter does, but I think I get the gist of it; and after testing the new version of the produced HTML, it does indeed seem to work well both on a big-screen device and on a mobile phone.

PR #86, opened 2022-02-14
Make build output belong to user who starts build

JLouisKaplan-Arm:
* Ensures that build output using
  the docker build script belongs
  to the user who triggered the build.

* Achieves this by adding a docker user
  with same uid and uname as the user who
  starts the docker build process and activates
  this user by default for the entrypoint.

### DIFF
--- a/build_with_docker.sh
+++ b/build_with_docker.sh
@@ -1,5 +1,4 @@
 #!/bin/sh
-readonly script_path="$0"
-readonly uid="$(stat -c "%u" "${script_path}")"
-readonly gid="$(stat -c "%g" "${script_path}")"
+readonly uid="$(id -u)"
+readonly gid="$(id -g)"
 docker build -t llsoftsecbook_build docker && docker run --rm --user="${uid}":"${gid}" --mount type=bind,source="$(pwd)",target=/src llsoftsecbook_build all
--- a/build_with_docker.sh
+++ b/build_with_docker.sh
@@ -1,3 +1,1 @@
-build_args="--build-arg uname=${USER} --build-arg uid=$(id -u)"
-docker build $build_args -t llsoftsecbook_build docker
-docker run --rm --mount type=bind,source="$(pwd)",target=/src llsoftsecbook_build all
+docker build -t llsoftsecbook_build docker && docker run --rm --mount type=bind,source="$(pwd)",target=/src llsoftsecbook_build all
--- a/build_with_docker.sh
+++ b/build_with_docker.sh
@@ -1,4 +1,4 @@
-readonly base="$(dirname "${BASH_SOURCE[0]}")"
-readonly uid="$(stat -c "%u" $base)"
-readonly gid="$(stat -c "%g" $base)"
+readonly script_path="${BASH_SOURCE[0]}"
+readonly uid="$(stat -c "%u" $script_path)"
+readonly gid="$(stat -c "%g" $script_path)"
 docker build -t llsoftsecbook_build docker && docker run --rm --user=${uid}:${gid} --mount type=bind,source="$(pwd)",target=/src llsoftsecbook_build all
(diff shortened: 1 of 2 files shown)


JLouisKaplan-Arm:
Hi @sam-ellis , I'm not able to request a review from you directly from the UI box for some reason, so tagging you here instead in case you want to have a look.

JLouisKaplan-Arm:
update: I figured out a much simpler workaround based on a variation of method (2) above that was not well documented. Will add a commit later tonight, but adding a comment now in case anyone tries to review the current implementation in the meantime.

JLouisKaplan-Arm:
The latest version just uses the `docker run` flag `--user`. Turns out you can specify a group that hasn't been created in the container. 

Current patch sets the output to belong to the same user and group as the build script itself.

In an earlier iteration, I had the output files' group set to the group with same name as the user ([created by default in Debian](https://wiki.debian.org/UserPrivateGroups)) but after thinking about various use cases think it probably makes more sense as it is now.

Look forward to feedback. Cheers

kbeyls:
For some reason, I'm not experiencing the issue you're describing (I'm using OSX as a development platform).
I like that with the latest changes you're fixing the issue you're seeing with a small modification.
It does seem however that you're using bash-specific functionality in the build_with_docker.sh script?
Therefore, I'd add a "#!/bin/bash" as the first line of the script, so that on systems/users where bash is not the default shell, the script does not break.

JLouisKaplan-Arm:
> For some reason, I'm not experiencing the issue you're describing (I'm using OSX as a development platform).

I looked into this abit more and it seems possibly related to how the Docker file sharing system for mac  (`osxfs`) works:

https://stackoverflow.com/a/61009413
https://docker-docs.netlify.app/docker-for-mac/osxfs/

I don't have access to a mac box for further exploration though unfortunately.

> It does seem however that you're using bash-specific functionality in the build_with_docker.sh script?

Good point, I didn't think about platform compatibility at all. I have changed `${BASH_SOURCE[0]}` to `$0` to make it Posix-shell compliant with only the loss of supporting a few unlikely usage edge cases: https://stackoverflow.com/a/35006505

Based on my googling, it looks like OSX machines should have access to `stat` but if you could confirm that would be great!

As for the rest of the script, I ran it through [ShellCheck](https://www.shellcheck.net/) with a `#!/bin/sh` header, and have made the suggested changes, apart from [this one](https://github.com/koalaman/shellcheck/wiki/SC2155#problematic-code-in-the-case-of-readonly) which would make things more verbose.

Happy to revert the change to `$0`, or fix the `readonly` parts (or remove `readonly` entirely) if you think it's worthwhile though.

(On a related note, is it potentially worth thinking about which platforms we want to commit to supporting longer term for the build system? Might be excessive, but thought it worth mentioning in case we want a Docker build script for CMD/Powershell at some point.)

kbeyls:
> > For some reason, I'm not experiencing the issue you're describing (I'm using OSX as a development platform).
> 
> I looked into this abit more and it seems possibly related to how the Docker file sharing system for mac (`osxfs`) works:
> 
> https://stackoverflow.com/a/61009413 https://docker-docs.netlify.app/docker-for-mac/osxfs/
> 
> I don't have access to a mac box for further exploration though unfortunately.
> 
> > It does seem however that you're using bash-specific functionality in the build_with_docker.sh script?
> 
> Good point, I didn't think about platform compatibility at all. I have changed `${BASH_SOURCE[0]}` to `$0` to make it Posix-shell compliant with only the loss of supporting `source`[ing the script](https://stackoverflow.com/a/35006505).
> 
> Based on my googling, it looks like OSX machines should have access to `stat` but if you could confirm that would be great!

I tried the current version of the PR on my macbook. Indeed stat is available, but it does not seem to understand the -c option. I'm getting:

`stat: illegal option -- c
usage: stat [-FlLnqrsx] [-f format] [-t timefmt] [file ...]
`

That is the only error message I see. Not sure if that -c is necessary?
Despite that error message being printed, I still see the book being built correctly both for the PDF and HTML versions.

> (On a related note, is it potentially worth thinking about which platforms we want to commit to supporting longer term for the build system? Might be excessive, but thought it worth mentioning in case we want a Docker build script for CMD/Powershell at some point.)

Fair point. I'm hoping this script is the only script where we'll have to worry about differences on host systems, as all other build processing should occur within a docker instance...

kbeyls (approved):
LGTM, I also tested this on OSX.

PR #87, opened 2022-02-15
Add CPU cache architecture section

kbeyls:
This is incrementally adding content to be able to explain cache covert channels

### DIFF
--- a/book.md
+++ b/book.md
@@ -406,7 +406,7 @@ but the main characteristics that are important to set up a covert channel tend
 to be similar across most popular implementations.
 
 Caches are small and much faster memories than the main memory that aim to keep
-a copy of the data at the most frequently accessed addresses in main memory. The
+a copy of the data at the most frequently accessed main memory addresses. The
 set of addresses that are used most frequently changes quickly over time as a
 program executes. Therefore, the addresses that are present in CPU caches also
 evolve quickly over time. The content of the cache may change with every
@@ -415,32 +415,32 @@ executed read or write instruction.
 On every read and write instruction, the cache micro-architecture looks up if
 the data for the requested address happens to be present in the cache. If it is,
 the CPU can continue executing quickly; if not, dependent operations will have
-to wait until the data returns from the much more slowly main memory.
-Typical access times is 3 to 5 CPU cycles for the fastest cache on a CPU versus
+to wait until the data returns from the much slower main memory.
+Typical access time is 3 to 5 CPU cycles for the fastest cache on a CPU versus
 hundreds of cycles for a main memory access.\index{memory access time}
 
-Most systems have multiple level of cache\index{multi-level cache}, each with a
+Most systems have multiple levels of cache\index{multi-level cache}, each with a
 different trade-off between cache size\index{cache size} and access
 time\index{cache access time}. Some typical characteristics might be:
 
-* A L1 (level 1) cache, 32kB in size, with an access time of 4 cycles.
-* A L2 cache, 256Kb in size, with an access time of 10 cycles.
-* A L3 cache, 16MB in size, with an access time of 40 cycles.
-* Main memory, gigabytes in size, with an access time of 100 cycles.
+* L1 (level 1) cache, 32kB in size, with an access time of 4 cycles.
+* L2 cache, 256Kb in size, with an access time of 10 cycles.
+* L3 cache, 16MB in size, with an access time of 40 cycles.
+* Main memory, gigabytes in size, with an access time of more than 100 cycles.
 
-When the data was not present in the cache, it typically is stored in the cache
-after it's fetched from a slower cache level or main memory. This is often a
-good decision to make as there's a high likelihood the same address will be
-accessed by the program soon after. This high likelihood is known as the
+If data is not already present in a cache layer, it is typically stored there
+after it has been fetched from a slower cache level or main memory. This is
+often a good decision to make as there's a high likelihood the same address will
+be accessed by the program soon after. This high likelihood is known as the
 [principle of locality](https://en.wikipedia.org/wiki/Locality_of_reference)\index{principle
 of locality}\index{locality of reference}.
 
 Data is stored and transferred between cache levels in blocks of aligned memory.
-Such a block is called a \index{cache block} or \index{cache line}. Typical
-sizes are 32, 64 or 128 bytes per cache line.
+Such a block is called a cache block\index{cache block} or cache
+line\index{cache line}. Typical sizes are 32, 64 or 128 bytes per cache line.
 
 When data that wasn't previously in the cache needs to be stored in the cache,
-most of the time, room has to be made for it be removing, or
+most of the time, room has to be made for it by removing, or
 evicting\index{cache eviction}, some other address/data from it. How that choice
 gets made is decided by the
 [cache replacement policy](https://en.wikipedia.org/wiki/Cache_replacement_policies)\index{cache
@@ -453,14 +453,15 @@ line; and pseudo-LRU approximates choosing the least recently used line.
 If an address can be stored in all locations available in the cache, the cache
 is fully-associative\index{fully-associative cache}. Most caches are however not
 fully-associative, as it's too costly to implement. Instead, most caches are
-set-associative\index{set-associative cache}. In a set-associate cache, a
-specific address can only be stored in a few locations. If an address can
-potentially be stored in 2 locations, the cache is said to be 2-way set
-associative. If it can be stored in one of at most 4 locations, it's called
-4-way set associative, and so on. When an address can only be stored in one
-location in the call, the cache is said to be direct mapped\index{direct mapped
-cache}, rather than 1-way set associative. Typical values of set associativity
-are direct mapped, 2-way, 4-way, 8-way, 16-way or 32-way set associative.
+set-associative\index{set-associative cache}. In an N-way set-associative cache,
+a specific main memory address can only be stored in one of N cache locations.
+For example, if an address can potentially be stored in one of 2 locations, the
+cache is said to be 2-way set-associative. If it can be stored in one of 4
+locations, it's called 4-way set-associative, and so on. When an address can
+only be stored in one location in the cache, it is said to be
+direct-mapped\index{direct-mapped cache}, rather than 1-way set-associative.
+Typical organizations are direct-mapped, 2-way, 4-way, 8-way, 16-way or 32-way
+set-associative.
 
 \missingcontent{Explain indexing mechanism used; from address bits to index in cache.}
 
--- a/book.md
+++ b/book.md
@@ -415,9 +415,9 @@ executed read or write instruction.
 On every read and write instruction, the cache micro-architecture looks up if
 the data for the requested address happens to be present in the cache. If it is,
 the CPU can continue executing quickly; if not, dependent operations will have
-to wait until the data returns from the much slower main memory.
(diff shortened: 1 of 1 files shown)


book.md
```diff
@@ -398,15 +398,76 @@ cache, without the processes having rights to read or write to the same memory
...
+evicts the cache line that is least recently used; random picks a random cache
+line; and pseudo-LRU approximates choosing the least recently used line.
+
+If an address can be stored in all locations available in the cache, the cache
+is fully-associative\index{fully-associative cache}. Most caches are however not
+fully-associative, as it's too costly to implement. Instead, most caches are
+set-associative\index{set-associative cache}. In a set-associate cache, a
+specific address can only be stored in a few locations. If an address can
+potentially be stored in 2 locations, the cache is said to be 2-way set
+associative. If it can be stored in one of at most 4 locations, it's called
```
JLouisKaplan-Arm, line 459: is the _at most_ true here? I.e. if a memory address could be stored at one of 3 locations I guess we wouldn't call it 4-way set associative?

also, I like the 'one of' here, it helped me understand things better. I wonder if the 2-way example should also have 'one of' just to be consistent, as I was initially wondering if somehow a 2-way cache might sometimes decide to store a memory address in two locations simultaneously.

book.md
```diff
@@ -398,15 +398,76 @@ cache, without the processes having rights to read or write to the same memory
...
+4-way set associative, and so on. When an address can only be stored in one
+location in the call, the cache is said to be direct mapped\index{direct mapped
```
JLouisKaplan-Arm, line 461: is '_call_' meant to be '_cache_'? Might just be my limited understanding here

book.md
```diff
@@ -398,15 +398,76 @@ cache, without the processes having rights to read or write to the same memory
...
+cache}, rather than 1-way set associative. Typical values of set associativity
```
JLouisKaplan-Arm, line 462: just noticing here that sometimes 'set-associative' is hyphenated and sometimes not in this paragraph, I'm not sure which convention is better to stick with though?

JLouisKaplan-Arm (commented):
I found this very clear and interesting! I learnt a lot reading it.

I have left some slightly pedantic and subjective comments and a few questions. I left the most comments/questions on the set-associative cache paragraph, which might be because this is the material that was most unfamiliar to me.

book.md
```diff
@@ -398,15 +398,76 @@ cache, without the processes having rights to read or write to the same memory
...
-### Typical cache architecture
-
-
-\missingcontent{Add a description of how a cache operates - as far as is
-necessary to explain the cache covert channel techniques.}
-
-### Flush + Reload
-
-### Prime + Probe
+### Typical CPU cache architecture
```
g-kouv, line 401: A general question would be whether this should be part of a special introductory chapter or even an appendix, since it's likely to be useful for other places too. Just a thought that can be skipped for now.

book.md
```diff
@@ -398,15 +398,76 @@ cache, without the processes having rights to read or write to the same memory
...
+
+There is a wide variety in
+[CPU cache micro-architecture](https://en.wikipedia.org/wiki/CPU_cache) details,
+but the main characteristics that are important to set up a covert channel tend
+to be similar across most popular implementations.
+
+Caches are small and much faster memories than the main memory that aim to keep
+a copy of the data at the most frequently accessed addresses in main memory. The
```
g-kouv, line 409: I think "addresses in main memory" is a little ambiguous here - "a copy of the data in main memory" could mean that either the copy or the data is in main memory. How about "main memory addresses" or "main memory locations"?

book.md
```diff
@@ -398,15 +398,76 @@ cache, without the processes having rights to read or write to the same memory
...
+executed read or write instruction.
+
+On every read and write instruction, the cache micro-architecture looks up if
+the data for the requested address happens to be present in the cache. If it is,
+the CPU can continue executing quickly; if not, dependent operations will have
+to wait until the data returns from the much more slowly main memory.
+Typical access times is 3 to 5 CPU cycles for the fastest cache on a CPU versus
+hundreds of cycles for a main memory access.\index{memory access time}
+
+Most systems have multiple level of cache\index{multi-level cache}, each with a
```
g-kouv, line 422: Typo: level -> levels

book.md
```diff
@@ -398,15 +398,76 @@ cache, without the processes having rights to read or write to the same memory
...
+
+When the data was not present in the cache, it typically is stored in the cache
+after it's fetched from a slower cache level or main memory. This is often a
+good decision to make as there's a high likelihood the same address will be
+accessed by the program soon after. This high likelihood is known as the
+[principle of locality](https://en.wikipedia.org/wiki/Locality_of_reference)\index{principle
+of locality}\index{locality of reference}.
+
+Data is stored and transferred between cache levels in blocks of aligned memory.
+Such a block is called a \index{cache block} or \index{cache line}. Typical
```
g-kouv, line 439: This is missing the words "cache block" and "cache line" in the rendered text. I was caught out by this too, so I'm wondering if we should change the way \index works. It would make uses like the one in the previous paragraph more tricky, though, unless we do something similar to what's done for links.

book.md
```diff
@@ -398,15 +398,76 @@ cache, without the processes having rights to read or write to the same memory
...
+sizes are 32, 64 or 128 bytes per cache line.
+
+When data that wasn't previously in the cache needs to be stored in the cache,
+most of the time, room has to be made for it be removing, or
+evicting\index{cache eviction}, some other address/data from it. How that choice
```
g-kouv, line 444: Nit: could evicting be emphasised, to highlight that you're introducing a common term here? If we decide to change \index, this could be done automatically.

book.md
g-kouv, line 461: direct mapped should probably be direct-mapped.

book.md
g-kouv: 'set-associative' is the common use, so I vote for that.

g-kouv, line 462: "Typical values of set associativity..." : I find this sentence a little awkward grammatically. But also a direct mapped cache is not set-associative. Perhaps this could be: "Typical cache organizations are..." WDYT?

g-kouv (commented):
It seems quite useful to introduce these pointers to common terminology, since I'm expecting you'll use them heavily in the sections to follow!

A few minor comments from me:

book.md
```diff
@@ -398,15 +398,76 @@ cache, without the processes having rights to read or write to the same memory
...
+set of addresses that are used most frequently changes quickly over time as a
+program executes. Therefore, the addresses that are present in CPU caches also
+evolve quickly over time. The content of the cache may change with every
+executed read or write instruction.
+
+On every read and write instruction, the cache micro-architecture looks up if
+the data for the requested address happens to be present in the cache. If it is,
+the CPU can continue executing quickly; if not, dependent operations will have
+to wait until the data returns from the much more slowly main memory.
+Typical access times is 3 to 5 CPU cycles for the fastest cache on a CPU versus
```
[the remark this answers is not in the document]
kbeyls: Probably best to change "access times" to "access time"?

book.md
```diff
@@ -398,15 +398,76 @@ cache, without the processes having rights to read or write to the same memory
...
+Most systems have multiple level of cache\index{multi-level cache}, each with a
+different trade-off between cache size\index{cache size} and access
+time\index{cache access time}. Some typical characteristics might be:
+
+* A L1 (level 1) cache, 32kB in size, with an access time of 4 cycles.
+* A L2 cache, 256Kb in size, with an access time of 10 cycles.
+* A L3 cache, 16MB in size, with an access time of 40 cycles.
+* Main memory, gigabytes in size, with an access time of 100 cycles.
+
+When the data was not present in the cache, it typically is stored in the cache
```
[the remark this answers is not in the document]
kbeyls: Thanks for the suggestion - I followed it in the updated patch I'll post soon.
Consistency between "it's" and "it is" is probably something we ought to aim for?

book.md
```diff
@@ -398,15 +398,76 @@ cache, without the processes having rights to read or write to the same memory
...
+replacement policy}. Popular replacement algorithms are Least Recently Used
+(LRU)\index{LRU replacement policy}, Random\index{random replacement policy} and
+pseudo-LRU\index{pseudo-LRU replacement policy}. As the names suggest, LRU
+evicts the cache line that is least recently used; random picks a random cache
+line; and pseudo-LRU approximates choosing the least recently used line.
+
+If an address can be stored in all locations available in the cache, the cache
+is fully-associative\index{fully-associative cache}. Most caches are however not
+fully-associative, as it's too costly to implement. Instead, most caches are
+set-associative\index{set-associative cache}. In a set-associate cache, a
```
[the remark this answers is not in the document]
kbeyls: Thanks for the suggestion - I'm going to follow it in the upcoming updated patch.

book.md
kbeyls: yep, should be cache.

book.md
kbeyls: thanks, I went for consistently using set-associative.

book.md
kbeyls: For now, I'm only aware of CPU cache structure being relevant to understand cache covert channels; which AFAIK are only relevant for the remote side-channel chapter.
If it turns out it's relevant for multiple chapters, we can "factor out" this section to somewhere else?

book.md
kbeyls: I decided to go for the suggested "main memory addresses".
I'm assuming some of the confusion may come from different people having a different mental model of what addresses are. Are they pointing to location of the abstract space of all memory locations in a running program (i.e. close to being a mathematical model)? Or are they the physical locations in DRAM chips? Or something else?
I hope the paragraph now becomes unambiguous enough that we won't have to go into "what is an address really?"....

book.md
kbeyls: Thanks!

I've raised #90 to capture the idea that we could use a better \index mechanism.

book.md
kbeyls: Maybe we should write up a writing convention in the appendix that when a term gets introduced it should be typeset in a particular way?

I've factored out this discussion into issue #91.

book.md
kbeyls: thanks!

book.md
kbeyls: Yeah - I went for "Typical organizations are..."

book.md
```diff
@@ -398,15 +398,77 @@ cache, without the processes having rights to read or write to the same memory
...
+set of addresses that are used most frequently changes quickly over time as a
+program executes. Therefore, the addresses that are present in CPU caches also
+evolve quickly over time. The content of the cache may change with every
+executed read or write instruction.
+
+On every read and write instruction, the cache micro-architecture looks up if
+the data for the requested address happens to be present in the cache. If it is,
+the CPU can continue executing quickly; if not, dependent operations will have
+to wait until the data returns from the much slower main memory.
+Typical access time is 3 to 5 CPU cycles for the fastest cache on a CPU versus
```
g-kouv, line 419: Nit/grammar: I'm missing an article here. "A typical access time.."

g-kouv (approved):
LGTM!

book.md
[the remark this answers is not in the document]
JLouisKaplan-Arm: I don't have a strong preference myself. I know we are following Chicago style, but I found this APA style blog post an interesting reference point:

https://blog.apastyle.org/apastyle/2015/12/contractions-in-formal-writing-whats-allowed-whats-not.html

My instinct would be to avoid contraction generally but not as a hard rule, as the style is somewhat formal but not completely academic?

PR #92, opened 2022-02-23
Enable adding diagrams and images to the book.

kbeyls:
This is an attempt to add support for adding diagrams/images to the book.


kbeyls:
The markdown-link-check failure indicates that the that checker does not seem to understand that pandoc automatically adds a .svg or .pdf extension. It'll need to be investigated how to work around/fix that.

contributing.md
```diff
@@ -104,6 +104,21 @@ communicate, please use [Discussions](<https://github.com/llsoftsec/llsoftsecboo
...
+https://github.com/jgraph/drawio-desktop/. Inkscape can also be used to draw
+them, but may be a bit more cumbersome to use for technical diagrams.
+
+We use SVG as the image source format. The sources are stored in the `img`
+sub-directory.
+
+For the HTML output, the SVG images are included as is.  For the PDF output,
+the SVGs are automatically converted to PDFs.  If you edit or add an SVG image,
+make sure to check the PDF output to see if the conversion worked as expected.
+A potential issue that can happen during conversion to PDF is explained at
```
g-kouv, line 119: Could you maybe say "common issue" instead of "potential issue"? My understanding is that it happens whenever any text is introduced, but this sentence initially made me think that this will be a rare issue.

g-kouv (approved):
This is most helpful, thank you! :)

I've only got a minor comment for the contributor guidelines.

contributing.md
kbeyls: That makes sense, I've now updated the PR accordingly

PR #96, opened 2022-03-22
Add suppport for pandoc-fignos filter

kbeyls:
This makes it possible to refer to figures from text.
See https://github.com/tomduck/pandoc-fignos for details.
This will be used by upcoming book content.


g-kouv (approved):
LGTM!

PR #108, opened 2022-06-29
Add section on code reuse attack mitigations

g-kouv:
Also amend some links in earlier sections

### DIFF
--- a/book.md
+++ b/book.md
@@ -142,13 +142,12 @@ attacks later evolved to [Return-Oriented Programming
 techniques.
 
 To defend against code reuse attacks, the [Address Space Layout Randomization
-(ASLR)](#code-reuse-attacks) and [Control-Flow Integrity
-(CFI)](#code-reuse-attacks) measures were introduced. \todo{Refine section
-links used here and in the previous paragraph.} This interaction between
-offensive and defensive security research has been essential to improving
-security, and continues to this day. Each newly deployed mitigation results in
-attempts, often successful, to bypass it, or in alternative, more complex
-exploitation techniques, and even tools to automate them.
+(ASLR)](#aslr) and [Control-Flow Integrity (CFI)](#cfi) measures were
+introduced. This interaction between offensive and defensive security research
+has been essential to improving security, and continues to this day. Each newly
+deployed mitigation results in attempts, often successful, to bypass it, or in
+alternative, more complex exploitation techniques, and even tools to automate
+them.
 
 Memory safe [@Hicks2014] languages are designed with prevention of such
 vulnerabilities in mind and use techniques such as bounds checking and
@@ -324,7 +323,7 @@ Some examples of how the attacker may use the obtained primitives:
 
 * Leak information, such as pointers to specific data structures or code,
   or the stack pointer.
-* Overwrite the stack contents, e.g. to perform a [ROP attack](#code-reuse-attacks).
+* Overwrite the stack contents, e.g. to perform a [ROP attack](#rop).
 * Overwrite non-control data, e.g. authorization state. Sometimes this
   step is sufficient to achieve the attacker's goal, bypassing the need for
   arbitrary code execution.
@@ -337,7 +336,7 @@ purposes of this chapter we will focus on:
 * Preventing memory vulnerabilities in the first place, thus stopping
   the attacker from obtaining powerful read/write primitives.
 * Mitigating the effects of read/write primitives, e.g. with mechanisms
-  to maintain [Control-Flow Integrity (CFI)](#code-reuse-attacks).
+  to maintain [Control-Flow Integrity (CFI)](#cfi).
 
 ## Stack buffer overflows
 
@@ -481,11 +480,11 @@ the attacker's wishes.
 
 Some of these vulnerabilities can be mitigated with the measures described in
 this section, but often more general measures to ensure memory safety or
-[Control-Flow Integrity](#code-reuse-attacks) are necessary. For example, in
-addition to the hardening of specific library functions, compilers can also
-implement automatic bounds checking for arrays where the array bound can be
-statically determined (`-fsanitize=bounds`), as well as various other
-"sanitizers". We will describe these measures in following sections.
+[Control-Flow Integrity](#cfi) are necessary. For example, in addition to the
+hardening of specific library functions, compilers can also implement automatic
+bounds checking for arrays where the array bound can be statically determined
+(`-fsanitize=bounds`), as well as various other "sanitizers". We will describe
+these measures in following sections.
 
 \todo{Rethink the later sections of this chapter and the order in which issues
 and mitigations are presented. The "non-control data exploits" section should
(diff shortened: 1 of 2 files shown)


book.md
```diff
@@ -680,7 +679,255 @@ signal handler.  The ability to fake a signal handler frame and call
...
+the argument points to the vtable of `A` or of `B`, which are stored
+consecutively and are the only allowed possibilities. The checks generated for
+different types of control-flow transfers are similar.
+
+[^cfi-flags]: The LTO and visibility flags are required by Clang's CFI.
+
+Another implementation of forward-edge CFI is Windows [Control Flow
+Guard](https://docs.microsoft.com/en-us/windows/win32/secbp/control-flow-guard),
+which only allows indirect calls to functions that are valid indirect
+control flow targets.
```
kbeyls, line 796: I wonder which functions are valid indirect control flow targets. Is it exactly the functions whose addresses are taken explicitly in source? Or something different? Or is that too complex to explain briefly here?

book.md
```diff
@@ -680,7 +679,255 @@ signal handler.  The ability to fake a signal handler frame and call
...
+point will be the same as when we signed the return address). If the PAC is
+correct, which will be the case in normal execution, the extension bits of the
+address are restored, so that the address can be used in the `ret` instruction.
+However, if the stored return address has been overwritten with an address with
+an incorrect PAC, the upper bits will be corrupted so that subsequent uses of
+the address (such as in the `ret` instruction) will result in a fault.  By
+making sure we don't store any return addresses without a PAC, we can
+significantly reduce the effectiveness of ROP attacks.
+
+Pointer Authentication can be used more widely, to also implement a
```
kbeyls, line 895: I'd say "widely, for example to also implement a ..."

book.md
```diff
@@ -680,7 +679,255 @@ signal handler.  The ability to fake a signal handler frame and call
...
+forward-edge CFI scheme, as is done in the arm64e ABI [@McCall2019].
+
```
kbeyls, line 897: I think it'd also be worthwhile to explicitly state that a lot of different CFI schemes can be implemented on top of the Pointer authentication instructions, citing the dozen or so academic papers, with a one or two-sentence description of what each does? This can be a TODO for a later patch though....

book.md
```diff
@@ -680,7 +679,255 @@ signal handler.  The ability to fake a signal handler frame and call
...
+### BTI
+
+[Branch Target Identification
+(BTI)](https://developer.arm.com/documentation/102433/0100/Jump-oriented-programming?lang=en)
+\index{BTI}, introduced in Armv8.5, offers coarse-grained forward-edge
+protection. With BTI, the locations that are targets of indirect branches have
+to be marked with a new instruction, `BTI`. There are a few different types of
```
kbeyls, line 904: I'd say "There are 3 different types of..."

book.md
```diff
@@ -680,7 +679,255 @@ signal handler.  The ability to fake a signal handler frame and call
...
+BTI instructions that permit different types of indirect branches (`BR`, `BLR`,
+or both).
```
kbeyls, line 906: I'd say at then end "(indirect jump, indirect call, or both)". I would hope more readers would understand that as not all readers might be very familiar with AArch64 assembly.

book.md
```diff
@@ -680,7 +679,255 @@ signal handler.  The ability to fake a signal handler frame and call
...
+
+With BTI, individual pages can be marked as guarded or unguarded.  Guarded
+pages are the ones where BTI applies. An indirect branch to a non-BTI or the
```
kbeyls, line 909: I would leave the fact that pages can be marked as guarded or unguarded for a later paragraph. This is about making deployment more feasible. Together with the no-op nature of BTI these are 2 details that helps with practical deployment. Maybe better to explicitly state that BTI has 2 specific features to help with deployment: marking pages as guarded or not and it being encoded in the hint space?

book.md
```diff
@@ -680,7 +679,255 @@ signal handler.  The ability to fake a signal handler frame and call
...
+wrong type of BTI instruction in a guarded page will raise a Branch Target
+Exception. The BTI instruction has been assigned to the hint space, therefore
+it will be executed as a no-op in cores that do not support BTI, aiding its
+adoption. Both Clang and GCC support generating BTI instructions, with the
+`-mbranch-protection=bti` flag, or, to enable both BTI and return address
+signing with Pointer Authentication, `-mbranch-protection=standard`.
+
+### Issues with CFI implementation
```
kbeyls, line 917: I'm wondering if "limitations", "weaknesses" or another word would be better than "Issues" here?

book.md
g-kouv: Technically, there are four. (Including BTI without j/c, which doesn't allow any of the above). I'll say four and add "or none" to the explanation.

book.md
g-kouv: Some of those papers are more general pointer integrity/memory safety measures, which would be more suitable in a later section. I will add a reference to at least a couple CFI-specific ones in this section now though, and a TODO to mention some more wherever relevant.

book.md
g-kouv: I don't think it's only address-taken functions, AFAIK this also includes exported library functions (based on https://connormcgarr.github.io/examining-xfg/). It probably includes C++ virtual functions too. But since this isn't described in the official documentation, I intentionally left it a bit vague. I think I should change this to say "are marked as valid control flow targets" rather than "are valid ...", though.

book.md
```diff
@@ -680,7 +679,255 @@ signal handler.  The ability to fake a signal handler frame and call
...
+  stp     x29, x30, [sp, #-16]!
+  mov     x29, sp
+  ldr     x8, [x0]                   // load vptr
+  adrp    x9, 400000 <_init-0x558>
+  add     x9, x9, #0x858
+  sub     x9, x8, x9
+  sub     x9, x9, #0x10
+  ror     x9, x9, #5
+  cmp     x9, #0x2
+  b.cs    4006ec <call_foo(A*)+0x38>
```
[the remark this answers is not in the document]
g-kouv: Oops, you're absolutely right, this makes no sense without the addresses.

book.md
```diff
@@ -680,7 +679,255 @@ signal handler.  The ability to fake a signal handler frame and call
...
+based on the value of the key and the value of the stack pointer (which at this
+point will be the same as when we signed the return address). If the PAC is
+correct, which will be the case in normal execution, the extension bits of the
+address are restored, so that the address can be used in the `ret` instruction.
+However, if the stored return address has been overwritten with an address with
+an incorrect PAC, the upper bits will be corrupted so that subsequent uses of
+the address (such as in the `ret` instruction) will result in a fault.  By
+making sure we don't store any return addresses without a PAC, we can
+significantly reduce the effectiveness of ROP attacks.
+
```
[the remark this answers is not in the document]
g-kouv: I'll elaborate as you recommend, and add a short mention of the fact that the probability of guessing a PAC depends on the number of available PAC bits, together with a short definition of pointer substitution attacks.

kbeyls (approved):
LGTM! Thank you!

PR #110, opened 2022-08-22
Add section on non-control data attacks

g-kouv:


book.md
```diff
@@ -959,7 +959,110 @@ code pointers: attacks against non-control data, which will be the topic of
...
+  authentication should be done, but simply serves to illustrate how a
+  buffer overflow into a non-control variable can have serious security
+  consequences.
+
+Figure @fig:non-control-data-attack shows the stack frame layout for this
+function when the code is compiled for AArch64 with Clang 10.0[^ymmv]. As the figure
+shows, an overflow of `passphrase` will overwrite `authenticated`, setting it
+to a non-zero value, even though the passphrase was incorrect.  The
+`authenticate` function will then return a non-zero value, incorrectly
+indication authentication success.
```
kbeyls, line 1007: s/indication/indicating/?

book.md
```diff
@@ -959,7 +959,110 @@ code pointers: attacks against non-control data, which will be the topic of
...
+attacks are a real issue, it leaves open a very important question: what are
+the limits of such attacks? It is tempting to assume that data-only attacks are
+somehow inherently limited, however it has been demonstrated in [@Hu2016] that
+they can, in fact, be very expressive[^caveat].
+
+[^caveat]: The authors describe how DOP gadgets can be chained to simulate
+  a Turing machine, making DOP attacks Turing-complete (it's not possible
+  to simulate the infinite tape of a Turing machine on any actual hardware,
+  of course). Turing-completeness is not, however, a particularly useful
+  measure of exploitability, as explained in [@Dullien2018].
```
kbeyls, line 1025: I wonder if the reason why Turing-completeness is not a useful measure of exploitability could be summarized here or elsewhere in the book? It seems like something interesting to know for most readers of the book?

book.md
```diff
@@ -959,7 +959,110 @@ code pointers: attacks against non-control data, which will be the topic of
...
+instructions and are reachable from memory errors, and how those gadgets can be
+stitched together with the help of dispatcher gadgets, the role of which is
+specifically to chain gadgets together.
+
+Stitching gadgets together is simpler for interactive attacks, where the
+attacker can keep providing malicious input to trigger the initial memory
+error and a certain chain of gadgets, as many times as needed. For
+non-interactive attacks, the MINDOP jump operations are required as well,
+used in conjunction with a memory location that provides a virtual program
+counter.
```
kbeyls, line 1045: Reading this, it made me think it was somewhat similar as the famous NSO attack (as described at https://googleprojectzero.blogspot.com/2021/12/a-deep-dive-into-nso-zero-click.html). If that NSO zero click is indeed another example of a DOP attack, I think it would be useful/interesting enough to refer to it as it has had a profound impact on the world. Not sure if we'd need to point to news articles that show how the NSO attacks were used in the real world to spy on e.g. journalists?

kbeyls (approved):
This is great! Thank you Georgia!
LGTM - I've just left a few comments of where I think some improvements are still possible.
Feel free to disagree with them!

book.md
JLouisKaplan-Arm, line 1007: indication -> indicating?

book.md
```diff
@@ -959,7 +959,110 @@ code pointers: attacks against non-control data, which will be the topic of
...
+![Stack frame for `authenticate`](img/non-control-data-attack){ width=60% #fig:non-control-data-attack }
+
+For many more simple examples of data-only attacks that can occur in real
+applications, see [@Chen2005]. Although this makes it clear that data-only
+attacks are a real issue, it leaves open a very important question: what are
+the limits of such attacks? It is tempting to assume that data-only attacks are
+somehow inherently limited, however it has been demonstrated in [@Hu2016] that
+they can, in fact, be very expressive[^caveat].
+
+[^caveat]: The authors describe how DOP gadgets can be chained to simulate
```
JLouisKaplan-Arm, line 1021: would it make sense to define the DOP acronym here rather than paragraph below?

JLouisKaplan-Arm (approved):
Really interesting and clear! Hope it's OK I just added two small nits. Thanks

book.md
g-kouv: I'm adding a sentence to explain this briefly.

book.md
g-kouv: Thanks! I'm reordering the text slightly to define DOP before the footnote.

book.md
g-kouv: Calling it a "DOP" attack doesn't seem right, as DOP is a specific technique that involves writing exploits in MINDOP. It does appear to be (for the largest part, anyway) a data-only attack, though. (The second part of its description, https://googleprojectzero.blogspot.com/2022/03/forcedentry-sandbox-escape.html, mentions explicitly how it doesn't attempt to get native code execution, ROP, JOP etc).

We already have the blog post in bibliography, so I'm adding an additional reference earlier on (where I mention very elaborate attacks), but as I'm not very confident about the nature of the attack past the initial stage described in the first blog, I'd rather not get into a lot of detail here right now.  I hope that's OK! We can add more details later, I think the best place would be either towards the start of the section (where I added the reference), or at the end, but separate from DOP.

PR #111, opened 2022-09-02
Combine all side channel content in a single chapter.

kbeyls:
While trying to write about cache covert channels and cache side-channels, it
became clear that trying to separate out covert channels (how can communication
happen through a covert channel?) from side-cchannels (how does a covert
channel get abused by an attacker?) doesn't work well. It makes the explanation
of them less natural and harder to both write and probably also consume by the
reader.

Let's combine all side channel content in a single chapter so that we don't
have to artificially separate describing covert and side-channels where that
results in a less clear description.

### DIFF
--- a/book.md
+++ b/book.md
@@ -1099,11 +1099,48 @@ first, it is called a side-channel attack. The entity not intending to
 communicate is called the victim\index{victim}. The other entity is sometimes
 called the spy\index{spy}.
 
-The rest of this chapter mostly describes a variety of common covert channel
-mechanisms. It does not aim to differentiate much on whether both ends intend
-to cooperate, or whether one end is a victim under attack of the other end.
+There are a lot of covert channels and side-channels. We describe the most
+important ones, and applicable mitigations, in this chapter.
 
-## Cache covert channels
+## Timing side-channels
+
+An implementation of a cryptographic algorithm can leak information about the
+data it processes if its run time is influenced by the value of the processed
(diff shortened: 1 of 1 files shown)


book.md
```diff
@@ -1099,11 +1099,48 @@ first, it is called a side-channel attack. The entity not intending to
 communicate is called the victim\index{victim}. The other entity is sometimes
 called the spy\index{spy}.
 
-The rest of this chapter mostly describes a variety of common covert channel
-mechanisms. It does not aim to differentiate much on whether both ends intend
-to cooperate, or whether one end is a victim under attack of the other end.
+There are a lot of covert channels and side-channels. We describe the most
```
JLouisKaplan-Arm, line 1102: is the use of `covert` here ok? Are the terms `covert channels` and `side channels` interchangable? I noticed you changed an instance of the former to the latter in line 1256. Might be a good opportunity to make it consistent in this paragraph but only if you think that makes sense from a technical perspective.

JLouisKaplan-Arm (approved):
lgtm! just added one question.

book.md
kbeyls: Thank you for the review and pointing this out!
I've searched for the term "covert" in the book and adapted text where it didn't make sense anymore.
There are still a few occurrences of "covert" left where I thought it made more sense to used the term "covert channel" than "side-channel".

g-kouv:
LGTM too!

PR #112, opened 2022-09-19
Describe cache side-channel mitigations.

kbeyls:


book.md
```diff
@@ -1319,8 +1319,76 @@ Collide+Probe\index{Collide+Probe}[@Lipp2020], etc.
...
+>     * variable
+
+
+#### Mitigations disallowing spy programs to share the cache with the victim program
+
+If the victim and the spy do not share a common channel -- in this case a cache
+level -- then a side channel cannot be created. Some typical ways to reduce the
+number of cache levels victim and spy share is disabling [cpu
+multithreading](https://en.wikipedia.org/wiki/Multithreading_(computer_architecture))\index{multithreading};
+making sure a spy cannot run in the same virtual machine of victim; etc.
```
JLouisKaplan-Arm, line 1365: I struggle to parse this last bit, but might just be me. Is it a list of three separated by semicolons?

1. disabling cpu multithreading
2. making sure a spy cannot run in the same virtual machine of a victim
3. etc

In which case I wonder if commas would be clearer, and just saying "For example, two ways" instead of "Some typical ways", so the etc. can be removed.

JLouisKaplan-Arm (approved):
Very interesting as always. Added two small comments - just personal preference so please ignore if you don't agree.

book.md
```diff
@@ -1319,8 +1319,76 @@ Collide+Probe\index{Collide+Probe}[@Lipp2020], etc.
...
+
+One could argue that
+[site isolation](https://developer.chrome.com/blog/site-isolation/)\index{site
```
g-kouv, line 1368: This could also use some additional information, as it relies on additional guarantees, e.g. by the OS.

book.md
```diff
@@ -1319,8 +1319,76 @@ Collide+Probe\index{Collide+Probe}[@Lipp2020], etc.
...
+into this category.
+
+#### Mitigations disabling the spy program to infer a cache status change in the victim program through its own cache status
+
+In some contexts, the resolution of the smallest time increment measurable by
+the spy program can be reduced so much that it becomes much harder to
+distinguish between a cache hit and a cache miss. Injecting noise and jitter
+into the timer also makes it harder to distinguish between a cache hit and cache
+miss. This is one of the mitigations in [javascript engines against Spectre
+attacks](https://v8.dev/blog/spectre)\index{Spectre}.
```
g-kouv, line 1379: It might be useful to link to https://developer.mozilla.org/en-US/docs/Web/API/Performance/now, in addition to the V8 blog.

book.md
```diff
@@ -1319,8 +1319,76 @@ Collide+Probe\index{Collide+Probe}[@Lipp2020], etc.
...
+
+Note that this is not always a perfect mitigation - there are often surprising
+ways that an attacker can get a fine-grained enough timer or use statistical
```
g-kouv, line 1382: Another good reference for constructing high-resolution timers where none are meant to be exposed is https://gruss.cc/files/fantastictimers.pdf.

g-kouv (approved):
LGTM overall! Some minor comments follow.

book.md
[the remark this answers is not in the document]
kbeyls: I've rewritten this section now. Could you have a look to see if is better now?

book.md
kbeyls: I've resolved this by referring to a separate section on site-isolation that I'm planning to write later as part of SpectreV1 mitigations.

book.md
kbeyls: Thanks! I've added a reference to that paper.

book.md
kbeyls: Thanks for that feedback! I've now rewritten this section, this sentence is no longer there.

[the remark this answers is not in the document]
g-kouv: Yes, looks good to me!

g-kouv:
LGTM!

PR #129, opened 2023-01-12
Use bootstrap for HTML output

kbeyls:
This is a first step towards moving to a much better HTML version of the book, see discussion in issue #32.

In this commit, the bare minimum has been done to recreate the existing layout with bootstrap. I've tried to add sufficient comments in the pandoc_template.html file so that it should be clear what each line does. I hope that will help with keeping the template file maintainable.

I see this is a first step on a longer road towards making a much nicer-looking HTML output, aiming for something similar to what Rust's mdbooks or bookdown.org books look like.

Note that by moving to bootstrap, the font seems to change a bit, and there is more spacing between lines. Not sure if we should consider trying to move back to the denser text from what we had before, or if we should stick with bootstrap's default.

Also, I made the "header" screen-wide. I.e., the left-hand side TOC does not appear at the top of the page, but under the header, i.e.  title, author, copyright info. We can change that back later if wanted.

Otherwise, I believe that the look-and-feel and functionality of the HTML output is the same as before.

- - -
This change is [Reviewable](https://reviewable.io/reviews/llsoftsec/llsoftsecbook/129)

### DIFF
--- a/theme/html/default.css
+++ b/theme/html/default.css
@@ -1,18 +1,5 @@
 /* Do not show bullet points for TOC entries: */
 #TOC ::marker { content: none; }
-/* Put the table-of-contents in 30%-ish wide left column. */
-#TOC {
-  position: fixed;
-  top: 0;
-  left: 0;
-  width: 28%;
-}
-/* Put the main text in a 50%-ish wide column next to the TOC. */
-body {
-  position: relative;
-  left: 30%;
-  width: 50%;
-}
 /* Use a sans-serif font: */
 body {
   font-family: sans-serif;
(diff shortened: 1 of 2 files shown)


kbeyls:
A zip containing the PDF and HTML version of the book with the PR can be downloaded here: https://github.com/llsoftsec/llsoftsecbook/suites/10332724077/artifacts/508685679 (this was built by the github checks above).

g-kouv (approved):
This looks great, Kristof, thanks! I like the slightly modified look and feel. I find the less dense text more readable.

> Reviewed 2 of 2 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/llsoftsec/llsoftsecbook/129#-:-NLevXfR5G7mmOzRI1av:b-wkod6j)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @JLouisKaplan-Arm)

JLouisKaplan-Arm (approved):
Looks great to me. Seems like a much better path moving forward than manual CSS. I agree with @g-kouv that the less dense text is more readable. Thank you!

> *[Reviewable](https://reviewable.io/reviews/llsoftsec/llsoftsecbook/129#-:-NLfhDRYC1v7OVqUlORB:bn7tqfn)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @kbeyls)

PR #136, opened 2023-01-20
[HTML] Fix scrollbar position on large screens

kbeyls:
The main content pane has class col-lg-8 by default.  When the TOC sidebar is visible, which has col-lg-4, this nicely divides the screen between the 2 columns.
However, when the TOC is made invisible, that leaves screen space unused (only 8 of the 12 bootstrap columsn are used), resulting in a scrollbar not appearing at the right hand side, but maybe at 2/3ths of the screen.

Fix this by adding/removing the col-lg-8 class from the main content based on TOC visibility/invisibility.

This commit also fixes a second thing on small screens.  Before, when the TOC was made visible, the main content pane also remained visible, albeit below the TOC. This could lead to confusing views with 3 scrollbars, and just not looking good. Therefore, for small screens always show either the TOC or the main content.

The complexity of the javascript is getting to the point where I'm wondering if for the TOC/main column logic maybe we would be better of going full-custom and not use bootstrap at all. I'm not sure at the moment, so just kept sticking to using bootstrap.

- - -
This change is [Reviewable](https://reviewable.io/reviews/llsoftsec/llsoftsecbook/136)

### DIFF
--- a/theme/html/pandoc_template.html
+++ b/theme/html/pandoc_template.html
@@ -132,26 +132,46 @@ $body$
   }
 
   jQuery(document).ready(function() {
+    let toc_sidebar = jQuery('#toc-sidebar');
+    let main_content_col = jQuery('#main-content-col');
+
+    function setTocVisibility(visible) {
+      if (visible) {
+        toc_sidebar.removeClass("d-none");
+        main_content_col.addClass("col-lg-8");
+        if (!isLG()) {
+          // on a small screen, make the main content invisible
+          // when the TOC is visible.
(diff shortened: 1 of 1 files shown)


JLouisKaplan-Arm (approved):
LGTM! Thanks!

> *[Reviewable](https://reviewable.io/reviews/llsoftsec/llsoftsecbook/136#-:-NMEKCn4CTEzxzoyB1T_:b3is8u8)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @kbeyls)

PR #141, opened 2023-01-26
[HTML] Make tooltips appear on the bottom of buttons

kbeyls:
... rather than to the left, so that a tooltip is less likely to hide other buttons.

However, for this first button ("toggle TOC"), this does not seem to be working: the tooltip continues to be shown on the right hand side, i.e. over the other buttons.

I could not find a way to change the behaviour.
Overall, this still is an improvement for the other buttons.

- - -
This change is [Reviewable](https://reviewable.io/reviews/llsoftsec/llsoftsecbook/141)

### DIFF
--- a/theme/html/pandoc_template.html
+++ b/theme/html/pandoc_template.html
@@ -56,19 +56,22 @@ $-- Implement the button row at the top of the screen.
      class="navbar navbar-light bg-light">
   <div class="container-fluid">
     <div> $-- div is needed to keep the buttons in it next to each other.
-    <button type="button"
+    <a role="button"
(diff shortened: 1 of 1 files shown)


JLouisKaplan-Arm (approved):
Thanks Kristof - seems like a good improvement!

> *[Reviewable](https://reviewable.io/reviews/llsoftsec/llsoftsecbook/141#-:-NMj_iSQ0lZOXA35mlK7:b40bw8)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @kbeyls)

PR #146, opened 2023-01-31
[HTML] Add a "report issue" button.

kbeyls:
I hope this will make it easier for readers to report issues they see in the content of the book. I hope it will also make it more likely for readers to report the issues they see.

- - -
This change is [Reviewable](https://reviewable.io/reviews/llsoftsec/llsoftsecbook/146)

### DIFF
--- a/theme/html/pandoc_template.html
+++ b/theme/html/pandoc_template.html
@@ -69,6 +69,12 @@ $-- Implement the button row at the top of the screen.
       data-bs-toggle="tooltip" data-bs-title="Github repo"
       href="$github-repo$">
       <i class="bi bi-github"></i></a>
+    <a id="open-issue-button"
+      class="btn btn-sm btn-outline-primary" role="button"
+      data-bs-placement="bottom"
+      data-bs-toggle="tooltip" data-bs-title="Report issue"
+      href="$github-repo$/issues/new?title=Issue%20summary&body=Your%20issue%20content%20here.">
+      <i class="bi bi-lightbulb"></i></a>
     <a id="view-source-button"
       class="btn btn-sm btn-outline-primary" role="button"
       data-bs-placement="bottom"


kbeyls (commented):
> *[Reviewable](https://reviewable.io/reviews/llsoftsec/llsoftsecbook/146#-:-NN6dCIv3S00gjrfegz5:btoi9ek)* status: 0 of 1 files reviewed, all discussions resolved (waiting on @g-kouv and @JLouisKaplan-Arm)

JLouisKaplan-Arm (approved):
The change itself looks good to me, though I can't decide if it's excessive to have both a GitHub and a 'report issue' button?

In favour of having it - not everyone is familiary with GitHub UI so it will be more intuitive.

Happy either way, let's see what @g-kouv thinks

> *[Reviewable](https://reviewable.io/reviews/llsoftsec/llsoftsecbook/146#-:-NN8vGBO-8DXfmKyA5Oy:bfn2930)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @g-kouv)

kbeyls:
> The change itself looks good to me, though I can't decide if it's excessive to have both a GitHub and a 'report issue' button?
> 
> In favour of having it - not everyone is familiary with GitHub UI so it will be more intuitive.

Adding this would result in having 3 github-related buttons: "go to repo", "suggest an edit" and "report an issue".
We could make this one "github" dropdown button, with the 3 actions in the dropdown.
However, I'm wondering if "hiding" the "report an issue" in a dropdown would result in a reader being slightly less likely to discover that there is an easy way to report an issue... If only I knew more about UX design...
An example of a project using a "github dropdown" is https://jupyterbook.org/en/stable/intro.html.

I'm also thinking that if there was a nice way to have a "report an issue/suggest an improvement" next to the text, that might be useful too. For example, what if we had the button appear in the sidenote area, once per section? I guess that would probably be too much. But maybe have it appear next to the "get link" when you hover over a section title might be an OK balance?

g-kouv (approved):
I'm also in favour of having the button, happy to keep it as it is and revisit later if we want to change it to a dropdown.

> I'm also thinking that if there was a nice way to have a "report an issue/suggest an improvement" next to the text, that might be useful too. For example, what if we had the button appear in the sidenote area, once per section? I guess that would probably be too much. But maybe have it appear next to the "get link" when you hover over a section title might be an OK balance?

I don't think I'd need/use additional buttons, since the existing buttons stay anchored when scrolling through the text, so I'm happy for this change to go in as-is! If we split the content into multiple files, individual "suggest an improvement" buttons might be more important.


:lgtm:

> *[Reviewable](https://reviewable.io/reviews/llsoftsec/llsoftsecbook/146#-:-NNGcP909nVvm3xoM9b6:b-ef1xd8)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @kbeyls)

PR #151, opened 2023-02-10
Improve timing side-channel discussion

### LINKED ISSUE #149: Concerning timing attacks
Concerning:

"(a) The control flow, i.e. the trace of instructions executed, does not change depending on the processed data. This guarantees that every time the algorithm runs, exactly the same sequence of instructions is executed, independent of the processed data.

At the moment, we do not know of a compiler implementation that actively helps to guarantee both (a) and (b). A great reference giving practical advice on how to achieve (a), (b) and more security hardening properties specific for cryptographic kernels is found in (Pornin 2018)."

There are implementations of automatic control-flow linearizers that ensure (a). For instance, we have worked on this implementation: http://cuda.dcc.ufmg.br/lif/, which runs on LLVM. The related paper is "https://homepages.dcc.ufmg.br/~fernando/publications/papers/LuigiCGO21.pdf", and the related repository is https://github.com/lac-dcc/lif/Your issue content here.

kbeyls:
... by referring to the work of Wu2018 and Soares2021.

This aims to fix #149.

- - -
This change is [Reviewable](https://reviewable.io/reviews/llsoftsec/llsoftsecbook/151)

### DIFF
--- a/book.bib
+++ b/book.bib
@@ -425,3 +425,21 @@ publisher = {Addison-Wesley Professional},
   pages     = {309--318},
   year      = {2012}
 }
+
+@InProceedings{Soares2021,
+  author    = {Luigi Soares and Fernando Magno Quintan Pereira},
+  booktitle = {2021 {IEEE}/{ACM} International Symposium on Code Generation and Optimization ({CGO})},
+  title     = {Memory-Safe Elimination of Side Channels},
+  year      = {2021},
+  publisher = {{IEEE}},
+  doi       = {10.1109/cgo51591.2021.9370305},
+}
+
+@InProceedings{Wu2018,
+  author    = {Meng Wu and Shengjian Guo and Patrick Schaumont and Chao Wang},
+  booktitle = {Proceedings of the 27th {ACM} {SIGSOFT} International Symposium on Software Testing and Analysis},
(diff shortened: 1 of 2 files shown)


pronesto (approved):
The new text looks good to me! Perhaps, at some point, it would be interesting to mention the Constantine approach of Borrello et al (https://github.com/pietroborrello/constantine). In constrast to the work of Wu and Soares, Constantine uses a dynamic analysis to: (i) find out the trip count of loops and (ii) find out the limits of buffers. (i) let's it deal with loops that cannot be fully unrolled at compilation time, and (ii) lets it avoid invalid memory accesses. On shortcoming with this approach is that it depends on having good inputs for the algorithm that will be linearized. These inputs should exercise the trip count of loops and cover the limits of the data structures used in the program. But, Constantine is implemented in LLVM (although it is not in the main distribution), and it has been drawing some attention from the research community. Also, the first author, Pietro Borrello, is pretty accessible, and is always willing to explain how his tool works.

kbeyls:
Thanks for the review!
I've added a todo to also explore and discuss https://github.com/pietroborrello/constantine in the book.

@all-contributors please add @pronesto for review

PR #207, opened 2023-06-29
Improve names for branch predictor components

kbeyls:
Aim to use general names rather than names for specific implementations. Based on feedback from
https://twitter.com/matt_dz/status/1672375841844518915

- - -
This change is [Reviewable](https://reviewable.io/reviews/llsoftsec/llsoftsecbook/207)

### DIFF
--- a/book.md
+++ b/book.md
@@ -1769,18 +1769,22 @@ most CPUs have multiple pieces of
 [branch predictor](https://en.wikipedia.org/wiki/Branch_predictor)\index{branch
 predictor} logic, such as:
 
-- A component that predicts the outcome of a conditional branch: taken or not
+- A component that predicts the outcome of a conditional
+  branch\index{Conditional branch direction predictor}: taken or not
   taken\index{taken branch}. The prediction is typically done based on the
   outcome of this and other branches in the recent past. In academic literature,
   this type of component is often called a Pattern History Table (PHT)
-  \index{Pattern History Table (PHT)}.
+  \index{Pattern History Table (PHT)}, referring to a specific implementation of
+  a conditional branch direction predictor.
 - A component that predicts the target of a taken branch, i.e. the address of
   the next instruction after a taken branch. In academic literature, such a
   component is often called a Branch Target Buffer (BTB) \index{Branch Target
-  Buffer}.
+  Buffer}, referring to a specific implementation of a branch target
+  predictor\index{branch target predictor}.
 - A component that is specialized to predict the next instruction after a
   function return instruction. Academic literature often calls this a Return
-  Stack Buffer (RSB) \index{Return Stack Buffer}.
+  Stack Buffer (RSB) \index{Return Stack Buffer}, referring to a specific
+  implementation of a return address predictor\index{return address predictor}.
 
 [^branch-prediction-performance]: Over time, new CPU designs tend to support
 having more instructions in flight. [@Eyerman2009, section 4.2.3] suggests that
@@ -1793,10 +1797,11 @@ push to increase the accuracy of branch predictors.
 A number of attacks have been described over the past few years. A few examples,
 categorized per branch predictor component they target, are:
 
-- PHT: BranchScope [@Evtyushkin2018], BlueThunder [@Huo2019]. These attacks
-  infer whether a branch is taken or not taken in a victim process. They do so
-  by carefully making sure that a branch in the spy process uses the same branch
-  predictor entry (i.e. aliases) with the targeted branch in the victim
+- Conditional branch direction predictor\index{Conditional branch direction
+  predictor}: BranchScope [@Evtyushkin2018], BlueThunder [@Huo2019]. These
+  attacks infer whether a branch is taken or not taken in a victim process. They
+  do so by carefully making sure that a branch in the spy process uses the same
+  branch predictor entry (i.e. aliases) with the targeted branch in the victim
   process. By measuring whether the branch in the spy process gets predicted
   correctly, one can derive whether the branch in the victim process was taken
   or not.
@@ -1808,17 +1813,19 @@ categorized per branch predictor component they target, are:
   can enable an attacker to derive the value of the secret key. These papers
   demonstrate deriving the secret key from implementations of specific
   cryptographic kernels. It can also be used to break ASLR\index{ASLR}.
-- BTB: SBPA [@Aciicmez2007], BranchShadow [@Lee2017]. These earlier attacks are
-  based on making a branch in the spy process alias in the BTB with a targeted
-  branch in the victim process. They use methods such as timing difference, last
-  branch records, instruction traces or performance counters to measure whether
-  the branch in the spy process caused a specific state change in the BTB.
-- RSB: Hyper-Channel [@Bulygin2008]. In this case, a spy process invokes $N$
-  calls to fill up the return stack predictor. Then it lets the victim process
-  execute. Then, the spy process can measure how many of its return stack
-  entries have been removed from the RSB, by measuring the number of $N$ returns
-  that get mis-predicted. If the number of calls in the victim process is
-  dependent on secret information, this could leak it.
+- Branch Target Predictor\index{Branch target predictor}: SBPA [@Aciicmez2007],
+  BranchShadow [@Lee2017]. These earlier attacks are based on making a branch in
+  the spy process alias in the BTB with a targeted branch in the victim process.
+  They use methods such as timing difference, last branch records, instruction
+  traces or performance counters to measure whether the branch in the spy
+  process caused a specific state change in the BTB.
+- Return address predictor\index{return address predictor}: Hyper-Channel
+  [@Bulygin2008]. In this case, a spy process invokes $N$ calls to fill up the
+  return stack predictor. Then it lets the victim process execute. Then, the spy
+  process can measure how many of its return stack entries have been removed
+  from the RSB, by measuring the number of $N$ returns that get mis-predicted.
+  If the number of calls in the victim process is dependent on secret
+  information, this could leak it.
 
 The papers referred to above contain detailed explanations of how they set up
 the attack. All of these attacks use a general 3-step approach, similar to
--- a/book.md
+++ b/book.md
@@ -1765,26 +1765,18 @@ loss.[^branch-prediction-performance] Therefore, most CPUs
 predict\index{predict} which instruction needs to be executed after a branch,
 before the branch instruction has completed. Correctly and quickly predicting
 the instruction after a branch instruction is so important for performance that
-most CPUs have multiple pieces of
-[branch predictor](https://en.wikipedia.org/wiki/Branch_predictor)\index{branch
-predictor} logic, such as:
+most CPUs have multiple
+[branch predictors](https://en.wikipedia.org/wiki/Branch_predictor)\index{branch
+predictor}, such as:
 
-- A component that predicts the outcome of a conditional
-  branch\index{Conditional branch direction predictor}: taken or not
-  taken\index{taken branch}. The prediction is typically done based on the
-  outcome of this and other branches in the recent past. In academic literature,
-  this type of component is often called a Pattern History Table (PHT)
-  \index{Pattern History Table (PHT)}, referring to a specific implementation of
-  a conditional branch direction predictor.
-- A component that predicts the target of a taken branch, i.e. the address of
-  the next instruction after a taken branch. In academic literature, such a
-  component is often called a Branch Target Buffer (BTB) \index{Branch Target
-  Buffer}, referring to a specific implementation of a branch target
-  predictor\index{branch target predictor}.
-- A component that is specialized to predict the next instruction after a
-  function return instruction. Academic literature often calls this a Return
-  Stack Buffer (RSB) \index{Return Stack Buffer}, referring to a specific
-  implementation of a return address predictor\index{return address predictor}.
+- A predictor of the outcome of a conditional branch\index{conditional branch
+  direction predictor}: taken or not taken\index{taken branch}. The prediction
+  is typically a history-based branch prediction, i.e., based on the outcome of
+  this and other branches in the recent past.
+- A predictor of the target of a taken branch\index{branch target predictor},
+  i.e. the address of the next instruction after a taken branch.
+- A predictor that is specialized to predict the next instruction after a
+  function return instruction.\index{return address predictor}
 
 [^branch-prediction-performance]: Over time, new CPU designs tend to support
 having more instructions in flight. [@Eyerman2009, section 4.2.3] suggests that
@@ -1797,7 +1789,7 @@ push to increase the accuracy of branch predictors.
 A number of attacks have been described over the past few years. A few examples,
 categorized per branch predictor component they target, are:
 
-- Conditional branch direction predictor\index{Conditional branch direction
+- Conditional branch direction predictor\index{conditional branch direction
   predictor}: BranchScope [@Evtyushkin2018], BlueThunder [@Huo2019]. These
   attacks infer whether a branch is taken or not taken in a victim process. They
   do so by carefully making sure that a branch in the spy process uses the same


kbeyls:
@MattPD: as this is a change based on your suggestion, are you interested to review this?

MattPD:
Sure, I'm going to take a look later today!

book.md
```diff
@@ -1769,18 +1769,22 @@ most CPUs have multiple pieces of
 [branch predictor](https://en.wikipedia.org/wiki/Branch_predictor)\index{branch
 predictor} logic, such as:
 
-- A component that predicts the outcome of a conditional branch: taken or not
+- A component that predicts the outcome of a conditional
+  branch\index{Conditional branch direction predictor}: taken or not
   taken\index{taken branch}. The prediction is typically done based on the
   outcome of this and other branches in the recent past. In academic literature,
   this type of component is often called a Pattern History Table (PHT)
```
MattPD, line 1776: I think I'd leave out PHT entirely as a specific implementation detail of a specific CBP design, but naturally feel free to decide whatever makes more sense in this context :-)

To provide some context on what I mean by a "specific design", the usage of PHT is inherently tied to two-level predictors [two-level BP which uses a branch history register (BHR) table and a branch pattern history table (PHT)]--cf. Section 3.1 in "A Survey of Techniques for Dynamic Branch Prediction", Concurrency and Computation: Practice and Experience (CCPE) Volume 31, Issue 1 (2019), Sparsh Mittal, https://arxiv.org/abs/1804.00261

Nowadays, geometric history length BPs (cf. Section 3.5) like TAGE (or TAGE-based) predictors also a common design with alternatives including Neural BPs (Section 7). Not all of these are PHT-based predictors (i.e., not all are using a single pattern history table; some brief notes on how the usage of history differs in tAGE in https://danluu.com/branch-prediction/, https://research.ac.upc.edu/80-in-2019/slides/Yale80-in-2019-Andre-Seznec.pdf), e.g., TAGE uses multiple predictor tables corresponding to several different history lengths (https://jilp.org/vol9/v9paper6.pdf). TAGE-based BPs have been implemented in Intel CPUs (e.g., Haswell), with AMD CPUs making use of perceptron (hashed perceptron predictor in Zen/Zen+; perceptron predictor _and_ TAGE predictor in Zen 2).

I think the reason PHT may show up often in academic literature is due to high influence of two-level BP design reflecting its historical importance (granted, to give credit where credit is due, it was a major achievement at the time).
For more recent designs cf. Section II "Current state of the art" in "Branch Prediction Is Not a Solved Problem: Measurements, Opportunities, and Future Directions", 2019, Chit-Kwan Lin, Stephen J. Tarsa, https://arxiv.org/abs/1906.08170. One example (from a particular core family) would be https://people.engr.tamu.edu/djimenez/pdfs/exynos_isca2020.pdf (Evolution of the Samsung Exynos CPU Microarchitecture, Section IV. Branch Prediction).

Again, make of that what you will :-) hopefully historical context may (or may not) be useful to make some sense of out it.

book.md
```diff
@@ -1769,18 +1769,22 @@ most CPUs have multiple pieces of
...
-  \index{Pattern History Table (PHT)}.
+  \index{Pattern History Table (PHT)}, referring to a specific implementation of
+  a conditional branch direction predictor.
 - A component that predicts the target of a taken branch, i.e. the address of
   the next instruction after a taken branch. In academic literature, such a
   component is often called a Branch Target Buffer (BTB) \index{Branch Target
-  Buffer}.
+  Buffer}, referring to a specific implementation of a branch target
```
MattPD, line 1782: In contrast, as far as branch target predictors are concerned, the usage of BTB is probably widespread (universal?) enough that I don't have any strong feelings on whether it's worth mentioning that it's just a specific implementation.

book.md
```diff
@@ -1769,18 +1769,22 @@ most CPUs have multiple pieces of
...
+  predictor\index{branch target predictor}.
 - A component that is specialized to predict the next instruction after a
   function return instruction. Academic literature often calls this a Return
-  Stack Buffer (RSB) \index{Return Stack Buffer}.
+  Stack Buffer (RSB) \index{Return Stack Buffer}, referring to a specific
```
MattPD, line 1786: "Return-Address Stack (RAS)" is another common term, possibly whether worth mentioning here.

book.md
```diff
@@ -1808,17 +1813,19 @@ categorized per branch predictor component they target, are:
...
+  They use methods such as timing difference, last branch records, instruction
+  traces or performance counters to measure whether the branch in the spy
+  process caused a specific state change in the BTB.
+- Return address predictor\index{return address predictor}: Hyper-Channel
+  [@Bulygin2008]. In this case, a spy process invokes $N$ calls to fill up the
+  return stack predictor. Then it lets the victim process execute. Then, the spy
+  process can measure how many of its return stack entries have been removed
+  from the RSB, by measuring the number of $N$ returns that get mis-predicted.
+  If the number of calls in the victim process is dependent on secret
+  information, this could leak it.
```
MattPD, line 1829: It may also be worth mentioning IBP (indirect branch predictor) used to predict indirect branches (e.g., C++ virtual calls).
// "indirect target array (ITA)" is another term, although AFAIK it's specific to AMD implementations.

book.md
kbeyls: Thanks for that elaborate feedback! My main reason to mention PHT here is to be able to refer back to it once I find the time to write about Spectre attacks. Even though the classification of Spectre attacks at https://transient.fail/ may not be perfect, I think it must be covered in the book, and they use names such as Spectre-PHT.
Granted, we can just as well introduce the term PHT there, rather than mentioning it here.

I'll think a bit more about this later today on what may be the best structure for the book.

MattPD: I see, I now understand where this is coming from--and also think that classification may be a good starting point, although I'd also take a look at some of the more recent ones (the authors and advisors do overlap, too, so I'm assuming this may reflect an evolution of how we can think about this, although it would be good to check/confirm that with the authors).

I think I'd consider slightly refining this paragraph to generalize it to be correct and useful either way, say:

```
"Correctly and quickly predicting the instruction after a branch instruction is so important for performance that
most CPUs have multiple [branch predictors](https://en.wikipedia.org/wiki/Branch_predictor)\index{branch
predictor}, such as:"

- A predictor of the outcome of a conditional branch\index{Conditional branch direction predictor}: taken or not  taken\index{taken branch}. The prediction is typically a history-based branch prediction, i.e., based on the outcome of this and other branches in the recent past. Microarchitectural implementations of history-based predictors can use a single Pattern History Table (PHT) or multiple PHTs.
```

(And similarly for the subsequent paragraphs, removing "component" from, e.g., "A component that predicts the target of a taken branch"; rationale: these aren't really components of a single branch predictor but rather distinct, separate predictors on their own.)

This way we introduce the notion of _conditional branch direction predictor_ itself, don't conflate it with a PHT (which it may or may not use--including the use of multiple PHTs), while explaining the commonly used acronym, too.

The reason I think this makes sense is that what's named "Spectre-PHT" nowadays affects most (all?) history-based branch predictors (some of which may use a PHT, some of which may use multiple PHTs, some of which may use a table or tables storing a hashed transformation of said history without it being an original, untransformed taken/untaken _pattern history_ per se).

FWIW, "branch predictor state" would make a good name for a more general class, used in https://arxiv.org/pdf/1807.03757.pdf, delineating the examples ("Branch predictor state – branch target buffer, branch history table, branch history register, RSB.").

I've been curious about the special treatment given to the PHT component, as the original https://spectreattack.com/spectre.pdf did note the distinction (similarly in Appendix A): "The structure of the branch history buffer (BHB) is a logical extension of the pattern history presented by [15]. The BHB helps make predictions on the basis of instruction histories, while preserving simplicity and the property of providing a rolling hash. 

The understanding of this distinction and the evolution of the BP implementations (including the usage of multiple PHTs) is also correct in the newer (2023) dissertation:

Remote Side-Channel Attacks and Defenses (2023)
https://martinschwarzl.at/media/files/thesis_main.pdf
- "Over the time, many different branch prediction schemes evolved, such as bi-mode branch prediction with 2 PHTs, index-sharing predictors, variable path length predictors, and perceptron-based predictors"
- "Intel CPUs have several branch predictors [21], e.g., the Branch History Buffer (BHB) [3, 34], Branch Target Buffer (BTB) [11, 34, 38], Pattern History Table (PHT)"
- "Spectre attacks [34] exploit speculative execution. Spectre-PHT [7] (also known as Spectre V1) exploits the Pattern History Table, which predicts the outcome of a conditional branch [34]. "
- "Branch prediction mechanisms are implemented via different structures, such as the Branch History Buffer (BHB) [6, 29], the Branch Target Buffer (BTB) [14, 29, 31], the Pattern History Table (PHT) [15, 29], and the Return Stack Buffer (RSB) [15, 30, 32]."

I think this 2022 paper has the best explanation of why we see such terminological... diversity :-) in practice (this paper is also a good reference as an example of the indirect branch predictors attacks):

Branch History Injection: On the Effectiveness of Hardware Mitigations Against Cross-Privilege Spectre-v2 Attacks
https://www.usenix.org/conference/usenixsecurity22/presentation/barberis

- "Even if the indirect branch predictor internals and naming widely change among different microarchitectures, the majority of them use a context-based approach [10, 15, 23, 34, 64]. For simplicity, we refer to BHB as a generic predictor structure that holds such context."

I was curious about the 2019 transient.fail classification for this, and in Section 2.2 it's "Among all the different variants, our work focuses on Branch Target Injection (BTI)—also known as Spectre v2 or Spectre-BTB [9]" while referring to "Branch Target Injection (BTI or Spectre v2)" in the abstract. Almost makes me wonder if "Spectre-v2" alone (dropping either PHT or BTB) isn't a better replacement?

What makes a clean classification (2019 transient.fail-style with a clear distinction between Spectre-PHT vs. Spectre-BTB vs. Spectre-RSB ) a bit tricky is that the resources being attacked and being subject to defenses may and in practice do overlap. E.g., Section 4.1.3. Branch prediction of "SoK: Analysis of Root Causes and Defense Strategies for Attacks on Microarchitectural Optimizations", https://arxiv.org/pdf/2212.10221.pdf, which does list a combination of resources like {R_BTB/R_PHT/R_BHB}.

Thinking out loud, then, I think going with a general paragraph (history-based conditional branch direction predictor) while mentioning PHT and its definition for historical completeness may make sense. Again, make of that what you will :-), I'm not extremely convinced one way or the other is clearly best (could be useful if we could get some of the authors to comment).

kbeyls (commented):
> *[Reviewable](https://reviewable.io/reviews/llsoftsec/llsoftsecbook/207)* status: 0 of 1 files reviewed, 4 unresolved discussions (waiting on @MattPD)

___
*[`book.md` line 1776 at r1](https://reviewable.io/reviews/llsoftsec/llsoftsecbook/207#gh-1247451753:-NZQvVB74iVxRoUr06OF:byb7l49) ([raw file](https://github.com/llsoftsec/llsoftsecbook/blob/a19b64c028c535b9aa2bc09bb56cc2dec9a6cae3/book.md#L1776)):*
<details><summary><i>Previously, MattPD (Matt) wrote…</i></summary><blockquote>

I see, I now understand where this is coming from--and also think that classification may be a good starting point, although I'd also take a look at some of the more recent ones (the authors and advisors do overlap, too, so I'm assuming this may reflect an evolution of how we can think about this, although it would be good to check/confirm that with the authors).

I think I'd consider slightly refining this paragraph to generalize it to be correct and useful either way, say:

```
"Correctly and quickly predicting the instruction after a branch instruction is so important for performance that
most CPUs have multiple [branch predictors](https://en.wikipedia.org/wiki/Branch_predictor)\index{branch
predictor}, such as:"

- A predictor of the outcome of a conditional branch\index{Conditional branch direction predictor}: taken or not  taken\index{taken branch}. The prediction is typically a history-based branch prediction, i.e., based on the outcome of this and other branches in the recent past. Microarchitectural implementations of history-based predictors can use a single Pattern History Table (PHT) or multiple PHTs.
```

(And similarly for the subsequent paragraphs, removing "component" from, e.g., "A component that predicts the target of a taken branch"; rationale: these aren't really components of a single branch predictor but rather distinct, separate predictors on their own.)

This way we introduce the notion of _conditional branch direction predictor_ itself, don't conflate it with a PHT (which it may or may not use--including the use of multiple PHTs), while explaining the commonly used acronym, too.

The reason I think this makes sense is that what's named "Spectre-PHT" nowadays affects most (all?) history-based branch predictors (some of which may use a PHT, some of which may use multiple PHTs, some of which may use a table or tables storing a hashed transformation of said history without it being an original, untransformed taken/untaken _pattern history_ per se).

FWIW, "branch predictor state" would make a good name for a more general class, used in https://arxiv.org/pdf/1807.03757.pdf, delineating the examples ("Branch predictor state – branch target buffer, branch history table, branch history register, RSB.").

I've been curious about the special treatment given to the PHT component, as the original https://spectreattack.com/spectre.pdf did note the distinction (similarly in Appendix A): "The structure of the branch history buffer (BHB) is a logical extension of the pattern history presented by [15]. The BHB helps make predictions on the basis of instruction histories, while preserving simplicity and the property of providing a rolling hash. 

The understanding of this distinction and the evolution of the BP implementations (including the usage of multiple PHTs) is also correct in the newer (2023) dissertation:

Remote Side-Channel Attacks and Defenses (2023)
https://martinschwarzl.at/media/files/thesis_main.pdf
- "Over the time, many different branch prediction schemes evolved, such as bi-mode branch prediction with 2 PHTs, index-sharing predictors, variable path length predictors, and perceptron-based predictors"
- "Intel CPUs have several branch predictors [21], e.g., the Branch History Buffer (BHB) [3, 34], Branch Target Buffer (BTB) [11, 34, 38], Pattern History Table (PHT)"
- "Spectre attacks [34] exploit speculative execution. Spectre-PHT [7] (also known as Spectre V1) exploits the Pattern History Table, which predicts the outcome of a conditional branch [34]. "
- "Branch prediction mechanisms are implemented via different structures, such as the Branch History Buffer (BHB) [6, 29], the Branch Target Buffer (BTB) [14, 29, 31], the Pattern History Table (PHT) [15, 29], and the Return Stack Buffer (RSB) [15, 30, 32]."

I think this 2022 paper has the best explanation of why we see such terminological... diversity :-) in practice (this paper is also a good reference as an example of the indirect branch predictors attacks):

Branch History Injection: On the Effectiveness of Hardware Mitigations Against Cross-Privilege Spectre-v2 Attacks
https://www.usenix.org/conference/usenixsecurity22/presentation/barberis

- "Even if the indirect branch predictor internals and naming widely change among different microarchitectures, the majority of them use a context-based approach [10, 15, 23, 34, 64]. For simplicity, we refer to BHB as a generic predictor structure that holds such context."

I was curious about the 2019 transient.fail classification for this, and in Section 2.2 it's "Among all the different variants, our work focuses on Branch Target Injection (BTI)—also known as Spectre v2 or Spectre-BTB [9]" while referring to "Branch Target Injection (BTI or Spectre v2)" in the abstract. Almost makes me wonder if "Spectre-v2" alone (dropping either PHT or BTB) isn't a better replacement?

What makes a clean classification (2019 transient.fail-style with a clear distinction between Spectre-PHT vs. Spectre-BTB vs. Spectre-RSB ) a bit tricky is that the resources being attacked and being subject to defenses may and in practice do overlap. E.g., Section 4.1.3. Branch prediction of "SoK: Analysis of Root Causes and Defense Strategies for Attacks on Microarchitectural Optimizations", https://arxiv.org/pdf/2212.10221.pdf, which does list a combination of resources like {R_BTB/R_PHT/R_BHB}.

Thinking out loud, then, I think going with a general paragraph (history-based conditional branch direction predictor) while mentioning PHT and its definition for historical completeness may make sense. Again, make of that what you will :-), I'm not extremely convinced one way or the other is clearly best (could be useful if we could get some of the authors to comment).
</blockquote></details>

Thank you for all your additional thoughts and references to papers.
I ended up deciding to remove all references to specific implementations from this section, as they are not needed here.
If and when references are needed in the to-be-written section on Spectre attacks: lets introduce them there.
___
*[`book.md` line 1829 at r1](https://reviewable.io/reviews/llsoftsec/llsoftsecbook/207#gh-1247462844:-NZQw6FwBBZKyLAEKJb6:b7kwuri) ([raw file](https://github.com/llsoftsec/llsoftsecbook/blob/a19b64c028c535b9aa2bc09bb56cc2dec9a6cae3/book.md#L1829)):*
<details><summary><i>Previously, MattPD (Matt) wrote…</i></summary><blockquote>

It may also be worth mentioning IBP (indirect branch predictor) used to predict indirect branches (e.g., C++ virtual calls).
// "indirect target array (ITA)" is another term, although AFAIK it's specific to AMD implementations.
</blockquote></details>

From the papers I've found and read specifically on "passive" branch predictor side-channels (i.e. not spectre-style attacks, which actively inject different state into branch predictors), I don't think I've seen any that target indirect branch predictors.
Therefore, I think I'll not mention them in this section.

MattPD (approved):

MattPD (approved):
Makes sense, LGTM!

kbeyls (commented):
> Reviewed 1 of 1 files at r2, all commit messages.
> Dismissed @MattPD from 3 discussions.
> *[Reviewable](https://reviewable.io/reviews/llsoftsec/llsoftsecbook/207)* status: all files reviewed, 1 unresolved discussion (waiting on @MattPD)

kbeyls (commented):
> Dismissed @MattPD from a discussion.
> *[Reviewable](https://reviewable.io/reviews/llsoftsec/llsoftsecbook/207)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @kbeyls)

PR #225, opened 2023-12-21
Update pandoc to recent version (3.0.1) so that we can use grid tables.

kbeyls:
pandoc-fignos is no longer supported on recent pandoc versions, so instead, this commit adds the simple lua script theme/fignos.lua that implements just the functionality that we are using. This results in the figure numbering in the PDF to no use chapter-based numbering (e.g. figure 2.3 for the 3rd figure in chapter 2), but rather uses linear numbering (figures are numbered from 1 to N throughout the whole document). This makes it consistent with how they are numbered in the HTML output, so could be argued to actually be an improvement.

Diffing the tex output before and after this patch shows that the only other difference is a minute difference in the links used for citations, which is presumably caused by using the pandoc built-in 'citeproc' rather than the deprecated pandoc-citeproc filter.

The HTML output is harder to diff. The older version of pandoc creates very long HTML lines, whereas the newer version of pandoc seemingly tries hard to make the HTML output fit within an 80-column width.

As far as I can see, there are no regressions.
The only differences I've spotted are the ones described above.

- - -
This change is [Reviewable](https://reviewable.io/reviews/llsoftsec/llsoftsecbook/225)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -66,7 +66,7 @@ build/index.html: build/book.html build
 build/book.tex: book.md book.bib Makefile build theme/tex/pandoc_template.tex \
 				theme/html/markup_issue.lua \
 				theme/html/markup_todo.lua \
-			    theme/fignos.lua \
+				theme/fignos.lua \
 				$(pdfimages)
 	pandoc $< -t latex \
 		--template theme/tex/pandoc_template.tex \
--- a/Makefile
+++ b/Makefile
@@ -10,15 +10,18 @@ PANDOCFLAGS = \
 	      --number-sections \
 	      --resource-path=. \
 	      --standalone \
-	      --self-contained \
-	      --filter pandoc-fignos \
-	      --filter pandoc-citeproc \
+	      --embed-resources \
 	      --metadata=VERSION:$(version)
 
+COMMONFILTERS = \
+          --lua-filter theme/fignos.lua \
+          --citeproc
+
 .PHONY: all clean pdf html
-all: pdf html default_pandoc_html_template default_pandoc_latex_template
+all: pdf html default_pandoc_html_template default_pandoc_latex_template native
 pdf: build/book.pdf
 html: build/book.html build/default.css build/index.html
+native: build/book.native
 
 # The source of images are in SVG format.
 # The below lines define to convert the SVG source images to PDF images such
@@ -42,6 +45,7 @@ build/book.html: book.md book.bib Makefile build theme/html/pandoc_template.html
 				 theme/html/convert_to_sidenote.lua \
 				 theme/html/markup_issue.lua \
 				 theme/html/markup_todo.lua \
+				 theme/fignos.lua \
 				 build/default.css $(svgimages)
 	pandoc $< -t html \
 		--template theme/html/pandoc_template.html \
@@ -51,7 +55,10 @@ build/book.html: book.md book.bib Makefile build theme/html/pandoc_template.html
 		--lua-filter theme/html/convert_to_sidenote.lua \
 		-M css=build/default.css \
 		--default-image-extension=svg \
-		-o $@ $(PANDOCFLAGS)
+		-o $@ $(PANDOCFLAGS) $(COMMONFILTERS)
+
+build/book.native: book.md book.bib Makefile build
+	pandoc $< -t native -o $@ $(PANDOCFLAGS)
 
 build/index.html: build/book.html build
 	cp build/book.html build/index.html
@@ -59,13 +66,14 @@ build/index.html: build/book.html build
 build/book.tex: book.md book.bib Makefile build theme/tex/pandoc_template.tex \
 				theme/html/markup_issue.lua \
 				theme/html/markup_todo.lua \
+			    theme/fignos.lua \
 				$(pdfimages)
 	pandoc $< -t latex \
 		--template theme/tex/pandoc_template.tex \
 		--default-image-extension=pdf \
 		--lua-filter theme/html/markup_issue.lua \
 		--lua-filter theme/html/markup_todo.lua \
-		-o $@ $(PANDOCFLAGS)
+		-o $@ $(PANDOCFLAGS) $(COMMONFILTERS)
 
 build/book.pdf: build/book.tex Makefile build
 	latexmk -pdf build/book.tex -output-directory=build
--- a/book.md
+++ b/book.md
@@ -31,6 +31,8 @@ header-includes:
     \stepcounter{TodoCounter}
     \oldtodo[inline,caption={\arabic{TodoCounter}. #1}]{\bcpanchant \textit{#1}}
   }
+  \usepackage{caption}
+  \captionsetup[figure]{labelformat=empty}
   ```
 ...
 
@@ -1695,8 +1697,8 @@ assume $N=12$ for a 12-way set-associative cache, the total cache size is
 $N*2^L*2^S=12*64*32=24$KB.](img/CacheIndexing){ width=100% #fig:cache-indexing}
 
 Specific bits in the memory address are used for different cache indexing
-purposes, as illustrated in +@fig:cache-indexing. The least-significant $L$
-bits, where $2^L$ is the cache line size, are used to compute an address's
+purposes, as illustrated in figure @fig:cache-indexing. The least-significant
+$L$ bits, where $2^L$ is the cache line size, are used to compute an address's
 offset within a cache line. The next $S$ bits, where $2^S$ is the number of
 cache sets, are used to determine which cache set an address maps to. The
 remaining top bits are "tag bits". They are stored alongside a line in the cache
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,7 +1,6 @@
 # SPDX-License-Identifier: CC-BY-4.0
 # SPDX-FileCopyrightText: © 2021 Arm Limited <kristof.beyls@arm.com>
 # syntax=docker/dockerfile:1
-#FROM ubuntu:23.10
 FROM debian:trixie
 # 1. install pandoc and texlive dependencies
 #    Also install git as we need that for computing version numbers.
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,18 +1,17 @@
 # SPDX-License-Identifier: CC-BY-4.0
 # SPDX-FileCopyrightText: © 2021 Arm Limited <kristof.beyls@arm.com>
 # syntax=docker/dockerfile:1
-FROM ubuntu:20.04
+#FROM ubuntu:23.10
+FROM debian:trixie
 # 1. install pandoc and texlive dependencies
 #    Also install git as we need that for computing version numbers.
 #    Also install librsvg2-bin to enable pandoc to convert SVG images to PDF.
 #    Also install python as the pandoc-fignos filter depends on it.
 RUN apt-get update && \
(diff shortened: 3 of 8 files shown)


JLouisKaplan-Arm (approved):
Looks good to me! Just added a few nits - prob not worth respinning the patch for unless you want to.

> Reviewed 1 of 6 files at r1.
> *[Reviewable](https://reviewable.io/reviews/llsoftsec/llsoftsecbook/225#-:-NmDqvYCDr5QYGEld_9V:barf7nj)* status: 1 of 6 files reviewed, 3 unresolved discussions (waiting on @g-kouv and @kbeyls)

___
*[`Makefile` line 69 at r1](https://reviewable.io/reviews/llsoftsec/llsoftsecbook/225#-NmDqMgM9zDFmpxhG6et:-NmDqMgM9zDFmpxhG6eu:b63bo91) ([raw file](https://github.com/llsoftsec/llsoftsecbook/blob/a3bdc8988d54e2d8bc8af9dd0422ac39b18cf6c7/Makefile#L69)):*
> ```Makefile
> 				theme/html/markup_issue.lua \
> 				theme/html/markup_todo.lua \
> 			    theme/fignos.lua \
> ```

nit - 1 less tab then lines above and below
___
*[`docker/Dockerfile` line 4 at r1](https://reviewable.io/reviews/llsoftsec/llsoftsecbook/225#-NmDqRCa6IIgnlApLcNK:-NmDqRCb7Z7mMr_5dW62:b5tljtw) ([raw file](https://github.com/llsoftsec/llsoftsecbook/blob/a3bdc8988d54e2d8bc8af9dd0422ac39b18cf6c7/docker/Dockerfile#L4)):*
> ```Dockerfile
> # SPDX-FileCopyrightText: © 2021 Arm Limited <kristof.beyls@arm.com>
> # syntax=docker/dockerfile:1
> #FROM ubuntu:23.10
> ```

nit - is the comment left in intentionally?
___
*[`docker/entrypoint.sh` line 1 at r1](https://reviewable.io/reviews/llsoftsec/llsoftsecbook/225#-NmDqcepCBP2AHEMsmzX:-NmDqcepCBP2AHEMsmzY:b-mffiu6) ([raw file](https://github.com/llsoftsec/llsoftsecbook/blob/a3bdc8988d54e2d8bc8af9dd0422ac39b18cf6c7/docker/entrypoint.sh#L1)):*
> ```Shell
> #!/bin/bash
> ```

nit - is bash needed here as everything looks posix shell friendly?

kbeyls (commented):
Thank you for finding those minor issues, they should all be fixed now.

In the mean while, I found another issue, related to the "clickable headers".
The § sign now shows up in the table of contents side-bar, see the screen shot below.



The table of contents is generated by the pandoc "macro" `$table-of-contents$`. It seems there is little configurability in how that can be generated.

In the version of pandoc we used, pandoc produces entries such as the following for TOC items:
```html
    <li><a href="#memory-vulnerability-based-attacks"><span class="toc-section-number">2</span> Memory vulnerability based attacks <span>§</span></a>
```

The new version of pandoc produces
```html
    <li><a href="#memory-vulnerability-based-attacks" id="toc-memory-vulnerability-based-attacks"><span class="toc-section-number">2</span> Memory vulnerability based attacks §</a>
```

Note that the `<span>` around `§` has disappeared.
I think we're currently using this CSS to make sure the `§` doesn't show up in the TOC sidebar:
```css
#toc-sidebar a span
{
display: none;
}
```
which obviously doesn't work anymore with the `span` no longer being there.

I think we'll need to go for a less hacky approach, which doesn't add `§` to the text and then tries to hide it using CSS. The approach taken in the pandoc documentation itself at https://pandoc.org/MANUAL.html seems to do the reverse: use CSS to add a marker purely from CSS. I think that doing what the pandoc documentation does may be the better (or at least more stable) approach as it's less likely to break.
It seems this snippet of CSS from view-source:https://pandoc.org/css/site.css?v=20231215194550 is the magic CSS that is most of what's needed:
```css
/* show the anchor links in headings if one hovers over the heading. */
.anchor:before {
  /* link symbol + variation selector-15 (UFE0E); the latter forces to use the
   current font, not some special emoji. */
  content: 'ðŸ”—\FE0E';
  display: inline-block;
  font-size: 14pt;
  left: -1.75em;
  line-height: 1.66;
  opacity: 0.2;
  padding-right: 0.5em;
  position: absolute;
}
.anchor:hover {
  text-decoration: none;
}
h1:hover > .anchor::before,
h2:hover > .anchor::before,
h3:hover > .anchor::before {
  opacity: 1;
}
/* end: heading anchor links */
```

> *[Reviewable](https://reviewable.io/reviews/llsoftsec/llsoftsecbook/225#-:-NmFh-Rg-sytUWoKg_nn:b-at00yq)* status: 0 of 6 files reviewed, all discussions resolved (waiting on @g-kouv and @JLouisKaplan-Arm)

___
*[`Makefile` line 69 at r1](https://reviewable.io/reviews/llsoftsec/llsoftsecbook/225#-NmDqMgM9zDFmpxhG6et:-NmFgNqu5w4MrZbBMwoy:bb3thr1) ([raw file](https://github.com/llsoftsec/llsoftsecbook/blob/a3bdc8988d54e2d8bc8af9dd0422ac39b18cf6c7/Makefile#L69)):*
<details><summary><i>Previously, JLouisKaplan-Arm (Jonathan Louis Kaplan) wrote…</i></summary><blockquote>

nit - 1 less tab then lines above and below
</blockquote></details>

Thanks for pointing out, should now be fixed.
___
*[`docker/Dockerfile` line 4 at r1](https://reviewable.io/reviews/llsoftsec/llsoftsecbook/225#-NmDqRCa6IIgnlApLcNK:-NmFgap90JvaUZpm-i5Q:b-us32kp) ([raw file](https://github.com/llsoftsec/llsoftsecbook/blob/a3bdc8988d54e2d8bc8af9dd0422ac39b18cf6c7/docker/Dockerfile#L4)):*
<details><summary><i>Previously, JLouisKaplan-Arm (Jonathan Louis Kaplan) wrote…</i></summary><blockquote>

nit - is the comment left in intentionally?
</blockquote></details>

Ah, a left-over from experimenting with how to best get an updated pandoc version.
It turns out that the latest ubuntu (23.10) still ships with a rather old pandoc version, which is why in the end I went with using debian:trixie.
___
*[`docker/entrypoint.sh` line 1 at r1](https://reviewable.io/reviews/llsoftsec/llsoftsecbook/225#-NmDqcepCBP2AHEMsmzX:-NmFgpy83bRvCeh4FW3V:b-igelz8) ([raw file](https://github.com/llsoftsec/llsoftsecbook/blob/a3bdc8988d54e2d8bc8af9dd0422ac39b18cf6c7/docker/entrypoint.sh#L1)):*
<details><summary><i>Previously, JLouisKaplan-Arm (Jonathan Louis Kaplan) wrote…</i></summary><blockquote>

nit - is bash needed here as everything looks posix shell friendly?
</blockquote></details>

Another left-over from experimenting (I think to try and make the pandoc-fignos filter still work, but that experiment failed). I reverted back to /bin/sh

JLouisKaplan-Arm:
That CSS looks like a nice approach to me, but I wonder how the pandocs manual manages to put all the body headers in the `anchor` CSS class? Can't quite remember how it works now or why I needed to add the hacky CSS, I think maybe pandocs filter's idea of a Header AST node is different from the HTML but I could be wrong.

kbeyls (commented):
It does seem like when pandoc filters run (e.g. the `clickable-headers.lua`), it of course runs on "pandoc Header"s. It also does seem that the HTML writer of pandoc which runs later does add quite a few things in the HTML header. So, even if we try to add an anchor at the start of the pandoc header, the HTML writer later adds other stuff such as section numbers before:
```html
<h2 data-number="2.2" id="exploitation-primitives"><span class="header-section-number">2.2</span> <a href="#exploitation-primitives" class="selflink"></a>Exploitation primitives</h2>
```

That of course results in anchors showing up not at the start of headers.
It seems pandoc works around this using the `position: absolute` CSS and reserving an entire column just for anchors at the top-level CSS page layout, IIUC?

In the just-added patch, I thought it would be simpler to just add the link symbol at the end, like you've originally done. The major difference in the patch is that the symbol is added using CSS `::after`, not by adding the text to the header. That results in the link symbol not showing up in the table of content. It seems like pandoc does remove all links in headers when making the TOC.

In more detail, the patch also switches to using a different symbol (the same as on the pandoc manual page) that might represent a link icon more clearly than the § symbol, that was used before this patch?

Also, the link icon is always shown, even when not hovering over the header, at low opacity. This has the advantage that when the title is just long enough to fit on a single line, but the link symbol doesn't, the reader sees the extra vertical white space after the header and before the first paragraph is for the link symbol. See the below screenshot for an example.


> *[Reviewable](https://reviewable.io/reviews/llsoftsec/llsoftsecbook/225#-:-NmluIPt0Z0rbiSV8-Nm:bz4r0su)* status: 0 of 8 files reviewed, all discussions resolved (waiting on @g-kouv and @JLouisKaplan-Arm)

JLouisKaplan-Arm (approved):
All looks good to me - just one question about the change to book.md. Approved in case this is intentional, but looks like a possible remnant of local testing?

> Reviewed 1 of 6 files at r1, 3 of 3 files at r2, 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/llsoftsec/llsoftsecbook/225#-:-NnhbKB458LAFT-X0R3M:b1fdykz)* status: 6 of 8 files reviewed, 1 unresolved discussion (waiting on @g-kouv and @kbeyls)

___
*[`book.md` line 1700 at r3](https://reviewable.io/reviews/llsoftsec/llsoftsecbook/225#-Nnhb2J1-lgGGUfPPSER:-Nnhb2J1-lgGGUfPPSES:b85n3ns) ([raw file](https://github.com/llsoftsec/llsoftsecbook/blob/170eee9d3248fc3fdc1c44e93a3fec637e3ff32d/book.md#L1700)):*
> ```Markdown
> 
> Specific bits in the memory address are used for different cache indexing
> purposes, as illustrated in figure @fig:cache-indexing. The least-significant
> ```

are the changes in this file intentional?

kbeyls (commented):
> *[Reviewable](https://reviewable.io/reviews/llsoftsec/llsoftsecbook/225)* status: 6 of 8 files reviewed, 1 unresolved discussion (waiting on @g-kouv and @JLouisKaplan-Arm)

___
*[`book.md` line 1700 at r3](https://reviewable.io/reviews/llsoftsec/llsoftsecbook/225#-Nnhb2J1-lgGGUfPPSER:-Nnhc7Av05awPsgpTL8K:b-njgs8q) ([raw file](https://github.com/llsoftsec/llsoftsecbook/blob/170eee9d3248fc3fdc1c44e93a3fec637e3ff32d/book.md#L1700)):*
<details><summary><i>Previously, JLouisKaplan-Arm (Jonathan Louis Kaplan) wrote…</i></summary><blockquote>

are the changes in this file intentional?
</blockquote></details>

Yes, this change is intentional.
The fignos.lua script that this patch added is a simple implementation to make `@fig:` references work with pandoc 3, now that we can no longer use pandoc-fignos. pandoc-fignos has support for a `+@fig:` syntax that automatically adds the word "Figure" in the text. The fignos.lua script does not support that feature (trying to keep the fignos.lua script simple). I think it's the right tradeoff to not support that feature in our own custom fignos.lua script. It's far easier to just write "figure" manually in the text...

PR #228, opened 2024-02-04
book.md: fix bib reference typo for array bounds post

kees:
Removes extra []s around the @Cook2023 bib reference so the URL will render correctly.

- - -
This change is [Reviewable](https://reviewable.io/reviews/llsoftsec/llsoftsecbook/228)

### DIFF
--- a/book.md
+++ b/book.md
@@ -1331,8 +1331,7 @@ will include bounds checks.
 
 Successfully using bounds checking compiler features for a large codebase
 requires substantial effort. An example of this is refactoring the Linux kernel
-to use bounds checks for flexible arrays, as described in [Kees Cook's
-blog]([@Cook2023]).
+to use bounds checks for flexible arrays, as described in [@Cook2023].
 
 There are also hardware-based mitigations for violations of spatial memory
 safety. For example,
--- a/book.bib
+++ b/book.bib
@@ -638,7 +638,7 @@ publisher = {Addison-Wesley Professional},
 
 @Electronic{Cook2023,
   author = {Kees Cook},
-  title  = {Bounded Flexible Arrays in C},
+  title  = {Bounded Flexible Arrays in {C}},
   url    = {https://people.kernel.org/kees/bounded-flexible-arrays-in-c},
   year   = {2023},
 }


book.md
```diff
@@ -1332,7 +1332,7 @@ will include bounds checks.
 Successfully using bounds checking compiler features for a large codebase
 requires substantial effort. An example of this is refactoring the Linux kernel
 to use bounds checks for flexible arrays, as described in [Kees Cook's
-blog]([@Cook2023]).
+blog](@Cook2023).
```
kbeyls, line 1335:
```suggestion
blog](https://people.kernel.org/kees/bounded-flexible-arrays-in-c).
```

kbeyls: Thank you for reporting this @kees !
When I tested locally, just removing the square brackets still resulted in a broken link.
Ultimately, I think the root issue here is mixing citation syntax (starting with `@` syntax) and the markdown syntax to create a regular link.
For this reference, I thought that probably it would be best to have a regular link, as the target is a blog post, rather than a formal citation. Not going with a citation also means that the reader gets taken to your blog post directly rather than having an indirection through the bibliography and having to click a second time to get to the blog post.

Does that seem allright to you?

kbeyls:
Oh - one more question: this project follows all-contributors, see https://github.com/llsoftsec/llsoftsecbook?tab=readme-ov-file#contributors- 
Would you be OK to be added there?

kees:
> Oh - one more question: this project follows all-contributors, see https://github.com/llsoftsec/llsoftsecbook?tab=readme-ov-file#contributors- Would you be OK to be added there?

Ah, cool! Yes, that would be fine. :)

book.md
kees: I was just trying to follow the examples in the rest of the doc. For example, the bib linkage for Aleph's stuff seems to work as expected. It seems without the linkage, the array-bounds blog post doesn't end up appearing at the end of the document either, so I think we should fix it to show up. I noticed others have comments like "leave this with a trailing comma" but I didn't understand what that was about. Is that needed here for some reason?

kees:
Okay, I've actually build tested this now. (I didn't have docker installed, but now I've got it all working.) This appears to be a proper fix now. (Commit has been updated.)

kbeyls:
@all-contributors please add @kees for code

book.md
kbeyls: I'm not sure where those "leave this with a trailing comma" comments come from in the book.bib file. Maybe @g-kouv knows? (Although don't expect an answer soon, I expect Georgia to be offline for a while).

I'm happy either way with having an entry in the bibliography or linking directly. Thanks for fixing this!
I'll be merging this soon.

kees: I experimented with this: a trailing comma means "do not order by surname". So I have no trailing comma and correctly appear as "Cook, Kees" but Aleph One does, and correctly appears as "Aleph One" in the references.

PR #254, opened 2024-07-18
Update bibliography for Dullien2020

BetaRays:
The previous DOI was broken, so I updated the citation data. (This also adds keywords.)

I checked with a quick shell script that other DOIs don’t return 404s, so this should be the only error.

- - -
This change is [Reviewable](https://reviewable.io/reviews/llsoftsec/llsoftsecbook/254)

### DIFF
--- a/book.bib
+++ b/book.bib
@@ -152,7 +152,8 @@ Please keep the comma after the author's name
   volume    = {8},
   number    = {2},
   pages     = {391-403},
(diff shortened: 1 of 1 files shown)


kbeyls:
Thank you for fixing this link and checking all the other DOI links!

kbeyls:
@all-contributors please add @BetaRays for code

PR #263, opened 2025-06-20
Add support for Example and Definitions; enable references them and section headers

kbeyls:
This patch makes it possible to have example divs and definitions divs.
It also contains the necessarily functionality to reference these example and definitions from text.
As part of this same patch, since the functionality is similar, it also enables referring to specific sections.

- - -
This change is [Reviewable](https://reviewable.io/reviews/llsoftsec/llsoftsecbook/263)

### DIFF
--- a/theme/fignos.lua
+++ b/theme/fignos.lua
@@ -2,27 +2,93 @@
 -- similarly to how https://github.com/tomduck/pandoc-fignos works.
 -- pandoc-fignos is no longer maintained, therefore, we re-implement
 -- the functionality of it that we depend on in our own filter here.
--- local logging = require 'theme/html/logging'
+-- Additionally, this filter also converts references to examples,
+-- definitions and section headers.
 
-figure_counter = 0;
-label2id = {};
+-- This filter follows roughly the following processing steps:
+-- 1. Figures are first-class objects in pandoc. Examples and definitions
+--    are divs with classes "example" and "definition".
+--    Their captions are stored in the "caption" attribute on the div.
+-- 2. The filter iterates through each figure, example, definition and section
+--    header and assigns a number (or hierarchical number, e.g. "2.1.3.5" for
+--    section headers) to each of them, and stores the number in the "count"
+--    attribute on the object.
+-- 3. - For figures, it adds "Figure x: " to the caption, with x being the
+--      number assigned in the previous step.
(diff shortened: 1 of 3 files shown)


kbeyls (commented):
Here are a few example that show what this script produces.

1. Example environment
  For the following markdown source code:
  ```
Probably the main reason why some CFI schemes put all C functions in a single
equivalence class, is that real-world C code bases often incorrectly implicitly
cast one C function pointer type to another. Example @ex:qsort-cfi shows an example.

::: {.example #ex:qsort-cfi 
     caption="C code mixing different function pointer types"}

`` `c
#include <stdlib.h>
int cmp_long(const long *a, const long *b) { return *a < *b; }
long sort_array(long *arr, long size) {
  /* The prototype of qsort is:
     void qsort(void *base, size_t nmemb, size_t size,
                int (*compar)(const void *, const void *)); */
  qsort(arr, size, sizeof(long), &cmp_long);
  return arr[0];
}
`` `

In this example, the function `cmp_long` has a different signature than the
function pointer type expected by `qsort`.

This code will run under CFI schemes that put all C functions in a single
equivalence class, but will fail under CFI schemes that partition C functions
based on their signature.

:::
```
  produces the following in the HTML output:

and the following in the Latex/PDF output:


> *[Reviewable](https://reviewable.io/reviews/llsoftsec/llsoftsecbook/263#-:-OTBqHNI7jPoBjB_yyRa:by3f6f2)* status: 0 of 3 files reviewed, 1 unresolved discussion

___
*[a discussion](https://reviewable.io/reviews/llsoftsec/llsoftsecbook/263#-OTBrQnE6vHc_ZSfYew_:-OTBrQnE6vHc_ZSfYewa:b-u0dgvb) (no related file):*
An example of a section header reference is:
```
A software-only implementation is the clang shadow stack, which is explained in
more detail in section @sec:clang-shadow-stack.

...

##### Clang Shadow Stack {#sec:clang-shadow-stack}

Clang also implements a backward-edge CFI scheme known as [Shadow
Stack](https://clang.llvm.org/docs/ShadowCallStack.html)\index{shadow stack}.
...
```

In the HTML output, this produces:


In the Latex/PDF output, this produces:

JLouisKaplan-Arm (approved):
LGTM.

(Reviewing directly in GitHub as `reviewable` did not seem to be working when I tried.)

PR #270, opened 2025-08-05
Copy-edit: pedantry about NULL versus NUL.

statham-arm:
'NULL' is the idiom in C for describing a null _pointer_: something that is the size of a machine word and used to indicate the absence of an address of memory elsewhere.

The term for the zero _character_ code, used to terminate strings in C, is 'NUL' (as defined by ANSI X3.4).

- - -
This change is [Reviewable](https://reviewable.io/reviews/llsoftsec/llsoftsecbook/270)

### DIFF
--- a/book.md
+++ b/book.md
@@ -620,15 +620,16 @@ fake call stack somewhere else, for example on the heap, and use a primitive
 that changes the stack pointer's value instead. This is known as stack
 pivoting\index{stack pivoting}.
 
-Note that this fake call stack contains NULL bytes, even without considering
+Note that this fake call stack contains zero bytes, even without considering
 the exact values of the various return addresses included. An overflow bug that
 is based on a C-style string operation would not allow an attacker to replace
 the stack contents with this fake call stack in one go, since C-style strings
-are null-terminated and copying the fake stack contents would stop once the
-first NULL byte is encountered. The ROP chain would therefore need to be
-adjusted so that it doesn't contain NULL bytes, for example by initially
-replacing the NULL bytes with a different byte and adding some more gadgets to
-the ROP chain that write zero to those stack locations.
+are [null-terminated](https://en.wikipedia.org/wiki/Null-terminated_string) and
+copying the fake stack contents would stop once the first zero byte is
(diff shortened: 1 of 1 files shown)


book.md
```diff
@@ -620,14 +620,14 @@ fake call stack somewhere else, for example on the heap, and use a primitive
 that changes the stack pointer's value instead. This is known as stack
 pivoting\index{stack pivoting}.
 
-Note that this fake call stack contains NULL bytes, even without considering
+Note that this fake call stack contains NUL bytes, even without considering
 the exact values of the various return addresses included. An overflow bug that
 is based on a C-style string operation would not allow an attacker to replace
 the stack contents with this fake call stack in one go, since C-style strings
 are null-terminated and copying the fake stack contents would stop once the
```
kbeyls, line 627: Thank you for taking the effort to correct this!
I'm assuming many readers might not be aware of the difference between NUL and NULL/null....
Maybe adding a link to https://en.wikipedia.org/wiki/Null-terminated_string might be worthwhile for those who want to know more?

Maybe putting the link on "null-terminated" might make most sense? :
```
are [null-terminated](https://en.wikipedia.org/wiki/Null-terminated_string) and copying...
```

statham-arm:
> I'd rather change the text to avoid C terminology (especially if unfamiliar to some readers), since ROP does not just apply to C code

Ah, I should have been clearer in the commit message! ANSI X3.64 isn't anything to do with C. It's the standard for ASCII. The name NUL (along with other well-known names for control characters like CR, LF, BS and ESC, plus many more obscure ones like CAN, SUB, STX, NAK, DCS etc) is defined in the ASCII character set.

But perhaps that makes it even more obscure. I agree that "zero byte" is an improvement.

g-kouv (approved):
LGTM, thanks!

> Reviewed 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/llsoftsec/llsoftsecbook/270#-:-OWubwnWEBJSc4Df5B2Q:b1bqdmn)* status: all files reviewed, 1 unresolved discussion (waiting on @statham-arm)

PR #273, opened 2025-08-20
Add an index mechanism that works for all formats, not just latex.

kbeyls:
This commit makes that work by changing how we indicate index entries in the main text.
So far, we used '\index{_text_to_add_to_index_}' in the main text, which is a latex-specific mechanism. To make this more format-neutral, this patch changes this to using spans of class "index" to indicate that some text needs to be added to the index. In this patch, it has the following features. More features can be added in later patches, if it shows they would be useful.

- An index entry is added by putting the text that should be indexed in a span with class "index". It looks like this: ... An [exploit primitive]{.index} is a mechanism ... This will add "exploit primitive" to the index. Before this patch, this would be written as: ... An exploit primitive\index{exploit primitive} is a mechanism ...
- When you want the entry in the index to be different from the text in the span, use the "entry" attribute on the span. For example: ... CFI schemes are sometimes classified as [coarse-grained]{.index entry="coarse-grained CFI"} or [fine-grained]{.index entry="fine-grained CFI"}. ...
- If you want multi-level index entries, use the same character as in latex to indicate the different levels. For example: ... [cache invalidation]{.index entry="cache!invalidation"} ... will add a sub-entry "invalidation" under the main entry "cache" in the index.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,6 +15,7 @@ PANDOCFLAGS = \
 
 COMMONFILTERS = \
           --lua-filter theme/fignos.lua \
+          --lua-filter theme/index.lua \
           --citeproc
 
 .PHONY: all clean pdf html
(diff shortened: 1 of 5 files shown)


JLouisKaplan-Arm:
LGTM. If I had to add a nitpick question: do you think it would be easier visually to separate each index link with some punctuation, maybe a comma? But this patch is more than ready to go in as-is; approved.

